PR #19, opened 2022-12-01
Add a nixos module for Geth

brianmcgee:


modules/geth.nix
```diff
@@ -0,0 +1,246 @@
...
+in
+{
+  ###### interface
+
+  options = {
+    services.ethereum-nix.geth = mkOption {
+      type = types.attrsOf (types.submodule gethOpts);
+      default = { };
+      description = mdDoc "Specification of one or more geth instances.";
+    };
```
brianmcgee, line 168: @aldoborrero I'm namespacing the services under `ethereum-nix` to avoid clashing with the existing service.

modules/geth.nix
```diff
@@ -0,0 +1,246 @@
...
+
+        extraArgs = mkOption {
+          type = types.listOf types.str;
+          description = mdDoc "Additional arguments passed to Go Ethereum.";
+          default = [ ];
+        };
+
+        package = mkOption {
+          default = selfPkgs.geth;
+          defaultText = literalExpression "pkgs.ethereum-nix.geth";
```
brianmcgee, line 153: @aldoborrero is this the correct way of referring to the the packages within `ethereum.nix`? See the top of this file where this file is basically a module factory.

modules/geth.nix
aldoborrero: @brianmcgee we can take advantage of using `disabledModule` option (as [you can see here](https://nixos.org/manual/nixos/stable/index.html#sec-replace-modules)). 

Although, having an explicit namespace makes this not clash with existing functionality. I would like to understand how for example, [home-manager does it](https://github.com/nix-community/home-manager/issues/2740).

modules/geth.nix
aldoborrero: Normally, I would use `pkgs.geth` as the user would configure the flake by using the overlay feature and, it can refer to it as I described before or with a clause like `inputs.ethereum-nix.packages.geth`.

I'm seeing at the top of the factory you're passing `{self, selfPkgs}` but, if you think carefully, there's no necessity of passing `selfPkgs` explicitly as we can use `self` instead.

Said that I think we should use the same approach as `home-manager` by creating [a suite of tests](https://github.com/nix-community/home-manager/tree/master/tests) to cover these scenarios.

aldoborrero (commented):
@brianmcgee good PR! I think it's valuable to provide a suite of tests that checks the module is configured correctly and that launches a geth instance. 

To that extent, I do remember @Mic92 his approach to performing such tests. Let's confirm with him what the best practices are.

Mic92:
@aldoborrero  are you talking about a nixos test? Here is a template for that: https://github.com/numtide/nixos-remote/blob/main/tests/from-nixos.nix https://github.com/numtide/nixos-remote/blob/main/tests/lib/test-base.nix

aldoborrero (approved):
LGTM

PR #40, opened 2022-12-08
feat: replace treefmt with treefmt-nix

brianmcgee:
* configures treefmt with nix instead of toml
* configures the formatter so you can run `nix fmt`
* updates Just to reflect the new format command


aldoborrero (approved):
LGTM

PR #78, opened 2023-01-20
feat/unify service args

brianmcgee:
- feat(modules/prysm): adapt service to use args instead of config file
- doc: update examples in README


aldoborrero (approved):
LGTM

PR #82, opened 2023-01-24
chore(erigon): update erigon 2.35.2 -> 2.36.1

selfuryon:
Another erigon update from 2.35.2 to 2.36.1.
Also tried to update via `nix-update`: it took 9m47s, dunno why so long


aldoborrero (approved):
LGTM.

I'll test with my local dev to see if it also happens to me.

PR #84, opened 2023-01-26
chore(lighthouse): update lighthouse v3.3.0 -> v3.4.0

selfuryon:
Hello!
Update lighthouse to v3.4.0 (and web3signer too)
But I have a question, why we need to rename `OUT_DIR` to `LIGHTHOUSE_WEB3SIGNER_BIN`? 
Should I make a PR to lighthouse to change that?

I also noticed that `OUT_DIR` is used in `testing/web3signer_tests/src/lib.rs` as a reference to builder script result, so I fixed that too


brianmcgee:
Thanks for the PR! I'll let @aldoborrero answer your questions above as he is the one who added Lighthouse.

aldoborrero (approved):
LGTM

aldoborrero:
@selfuryon I do remember I had to rename it as the code was trying to fetch the web3 binaries from an external repository. Maybe we can create directly a patch that doesn't perform that step and we can avoid having the variable renaming.

Issue #90, opened 2023-01-27
Add a service for Vouch


selfuryon:
I'll try to implement that too in the same way as other modules

aldoborrero:
I assigned this task to you @selfuryon !

Issue #89, opened 2023-01-27
Add a service for Dirk


selfuryon:
I'll try to implement that in the same way as other modules

PR #92, opened 2023-02-02
feat/service improvements

brianmcgee:
Simplify the service implementations, re-using common code between them and removing anything related to management of mounts which will be done separately in a new module. 

- chore: update flake.lock
- feat(modules): simplify services
- fix(modules): do not add client packages to system
- feat(modules): use ExecStart instead of script
- feat(modules): remove redundant options in base service config

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -29,11 +29,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1673362319,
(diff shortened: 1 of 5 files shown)


brianmcgee:
@selfuryon when implementing #89 and #90 you should base your new services on the new pattern in this PR.

aldoborrero (approved):
LGTM

selfuryon:
> @selfuryon when implementing #89 and #90 you should base your new services on the new pattern in this PR.

Got that, I will do, thanks!

PR #93, opened 2023-02-02
feat/state snapshot

brianmcgee:
- feat(modules): better organise modules
- feat(modules): namespace services under ethereum.<type> where type is consensus or execution
- feat(modules/snapshot): basic control of service state volume
- feat(modules): replace foldListToAttrs lib function with mkMerge
- feat(modules): basic volume setup and snapshotting
- feat(modules/snapshot): add ISO 8601 timestamp to snapshots
- feat(modules/snapshot): add restart interval for periodic snapshotting
- feat(modules): refine namespacing changes, removing the consensus and execution part of the key
- feat(modules/snapshot): update the README


aldoborrero (approved):
Good work.

From my point of view, I think we should start creating releases in `ethereum.nix` to incorporate the fact we may break things.

That way we can avoid having our consumers discover with our latest change we're breaking them their config (as we are still iterating and stabilizing how we would like to define our modules and snapshots and whatnot).

I'll add an issue to track this specific fact.

Otherwise LGTM.

brianmcgee:
I agree. There are also strategies for renaming options which we can start following, but given we are still in the early days I'd like to keep breaking things as necessary is in the short term.

PR #96, opened 2023-02-07
fix: move pkgs and lib to imports

aldoborrero:
This PR sets correctly `pkg` and `lib` for flake.parts by setting them on `imports` clause.


brianmcgee (approved):
LGTM

PR #106, opened 2023-02-15
Add arm darwin support

SpontaneousOverthrow:
Currently I need to use ethdo on apple silicon.
Tested with:
```
nix build ".#ethdo" 
nix run ".#ethdo" 
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -68,7 +68,10 @@
           ./packages
           ./modules
         ];
-        systems = ["x86_64-linux"];
+        systems = [
+          "x86_64-linux"
+          "aarch64-darwin"
+        ];
       })
     .config
     .flake;


brianmcgee:
@SpontaneousOverthrow adding an extra system like that is going to give the mistaken impression that all the other packages and apps have been tested to work on Darwin. If you run `nix flake check` from an `aarch64-darwin` system you'll see that Nethermind for example isn't supported. 

You can modify the `perSystem` entry in `packages/default` such that for `x86_64-linux` it includes everything and for `aarch64-darwin` it only includes `ethdo` for now until we have time to review what is supported and add them in.

SpontaneousOverthrow:
Is this correct that the best way is to use `withSystem` module, where I need to provide packages for each arch?
https://flake.parts/module-arguments.html#withsystem

brianmcgee:
> Is this correct that the best way is to use `withSystem` module, where I need to provide packages for each arch? [flake.parts/module-arguments.html#withsystem](https://flake.parts/module-arguments.html#withsystem)

I've been looking into this a bit more today and It's more complex than I first thought. I'm gonna keep playing around with this branch https://github.com/nix-community/ethereum.nix/tree/feat/multi-system but there are some issues I need to resolve.

selfuryon:
Maybe try to use meta attributes and `meta.platforms` and filter by it?

selfuryon:
Like [that](https://github.com/selfuryon/ethereum.nix/pull/4/files#diff-c553fb5f9d4af3d9e20b25c437e11c7e6c7b1e5d80f311eb53aa1bf96e62e8e3): just use this function:
```
  platformPkgs = system: packages:
    filterAttrs (name: value: elem system (value.meta.platforms or [])) packages;
```
and then use that in current attrset:
```
    packages = platformPkgs system rec {
       ...
    }   
 ```

brianmcgee:
> Maybe try to use meta attributes and `meta.platforms` and filter by it?

I had this same thought, will have a look a bit later in the week.

brianmcgee:
Closing in favour of #114, we can continue the discussion and refinements there

brianmcgee:
@SpontaneousOverthrow #114 has been merged. I've tested ethdo a bit on my own M1 mac but I haven't used it in anger. Lemme know if there's any issues.

PR #113, opened 2023-02-20
chore: removal of unnecessary nixpkgs specification for flake.parts

aldoborrero:
Removes unnecessary specification of `nixpkgs` in `flake.parts` (it's the default action inside the framework).


brianmcgee (approved):
LGTM

Issue #50, opened 2023-01-04
How best to document everything?


brianmcgee:
The readme is only getting bigger. We should have a more robust set of docs. I'm open to suggestions.

aldoborrero:
I think it's time to use either [mdbook](https://rust-lang.github.io/mdBook/) or [docsify](https://docsify.js.org/#/).

aldoborrero:
We will stick with `mdbook` as it's quite simplistic to use.

brianmcgee:
I'd like to throw [MkDocs](https://www.mkdocs.org/) into the mix, I see it around on a lot of nix projects.

brianmcgee:
After speaking with the Numtide guys they are standardising on mkdocs. I had a look and it seems to be feature rich. An added bonus is I can just copy some of the setup and layout from projects like [Treefmt](https://github.com/numtide/treefmt) which I've already done in #147

PR #123, opened 2023-02-23
Client backup and restore

### LINKED ISSUE #72: Allow specifying btrfs sub volumes in geth service
This will be a more general feature eventually, but to begin with let's focus on the geth service and btrfs in particular.

The basic idea is to allow a user to specify that the `datadir` should be a btrfs subvolume. 

Before starting the geth client, the volume should be created if it hasn't already, with default mount options of `["discard" "noatime" "nodatacow"]`. If it already exists then ideally we validate the mount options match but that's not necessary for a first version.

I think we can leverage https://github.com/nix-community/disko. 

@Mic92 @Lassulus any advice/recommended approaches for integrating the service in `modules/geth.nix` with disko to achieve the outcome above?

brianmcgee:
This removes the snapshot functionality and replaces it with backup and restore that have options which are mixed into the service options allowing it to be configured in place. 

Currently only `geth`, `prysm-beacon` and `nethermind` support backup and restore, with automated testing in place for only Geth at the moment. I have manually tested with Geth and Prysm however this PR is getting bigger and bigger. I want to get this merged and tackle the following up tasks in smaller steps such as #144 and #50 and start iterating based on production feedback.

There is a lot more to document than is being included in this PR description. I want to tackle that more thoroughly as part of #50.

## Backup

When enabled a few things happen:

1. A metadata service `{name}-metadata.service` is created (e.g. `geth-sepolia-metadata.service`) which is managed by `{name}-metadata.timer` which polls the client every 10 seconds for information related to chain height. This is stored within the backup metadata folder `.backup` inside the state directory. 
2. A backup service `{name}-backup.service` is created which is managed by `{name}-backup.service` and runs on the configured schedule with a default of every 24 hours.

There are two types of backups which can be performed when `{name}-backup.service` is triggered:

1. In-situ backup
2. BTRFS snapshot based backup

In the case of an in-situ backup the client service is stopped and it's state directory is backed up via [borgbackup](https://www.borgbackup.org/). When the backup is complete the service is restart.

For a BTRFS snapshot based backup the client service is restarted, forcing a btrfs read-only snapshot to be created in the configured snapshot directory. That directory is then traversed and each snapshot found is backed up if it has not already been. Any snapshots older than a configurable number of days is then deleted to prevent a large number of snapshots being built up. 

When a backup is being created a content hash is created which is later used by the restore process to ensure the integrity of a given backup. 

```nix
services.ethereum.geth-sepolia = {
	backup = {
    	enable = true;

      	# btrfs = {
        #	 enable = true;
        # 	 snapshotDirectory = "/ethereum/snapshots/geth-sepolia";
      	# };

      	borg = {
        	repo = "ssh://borg@dione/data/borgbackup/ethereum/geth-sepolia";
        	keyPath = sops.secrets.borg_backup_key.path;
        	encryption.mode = "none";
        	strictHostKeyChecking = false;	
        	unencryptedRepoAccess = true;
      	};

      schedule = "0/12:00:00";	# every 12 hours
    };
}
```

## Restore

When restore is enabled an `ExecStartPre` script is added to the client's systemd service which will pull a configured snapshot from a remote borg backup repository and unpack it into the service's state directory. This will only occur if the state directory is empty and if the content hash is correct after unpacking.

```nix
services.ethereum.geth-sepolia = {
	restore = {
    	enable = true;
		snapshot = "1785690";

      	borg = {
        	repo = "ssh://borg@dione/data/borgbackup/ethereum/geth-sepolia";
        	keyPath = sops.secrets.borg_backup_key.path;
        	strictHostKeyChecking = false;	
        	unencryptedRepoAccess = true;
      	};
    };
}

```

Closes #72


brianmcgee:
I'm going to go ahead and merge this as it is blocking follow up work and @aldoborrero doesn't have the time to review it today.

PR #130, opened 2023-02-27
feat: prefer nixpkgs stable

brianmcgee:
`nix fmt` was broken with a glibc error and I've been noticing issues with latest unstable across a few projects the last few days.

In general I think we should be building on the latest stable release and bringing in unstable only when necessary

### DIFF
--- a/packages/default.nix
+++ b/packages/default.nix
@@ -9,6 +9,7 @@
 
   perSystem = {
     self',
+    inputs',
     config,
     pkgs,
     system,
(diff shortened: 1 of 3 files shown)


packages/default.nix
```diff
@@ -9,17 +9,20 @@
...
+    inputs',
     config,
     pkgs,
     system,
     ...
   }: let
     inherit (pkgs) callPackage;
     inherit (lib.flake) platformPkgs platformApps;
+
+    callPackageUnstable = inputs'.nixpkgs-unstable.legacyPackages.callPackage;
```
brianmcgee, line 21: `lighthouse` is the only package which requires unstable at the moment

aldoborrero (approved):

PR #114, opened 2023-02-22
feat/multi system

brianmcgee:
- feat(packages): add meta.platforms for all packages
- feat(packages): add conditional inclusion of packages based on meta.platforms
- feat(lib): simplify lib construction
- feat(lib): move package related functions to lib
- feat(apps): system specific apps
- feat(flake): add aarch64-darwin system
- feat(overlay): simplify specification of overlayAttrs

PR #131, opened 2023-02-27
fix(apps): geth and prysm apps were broken by #114

brianmcgee:


aldoborrero (approved):
We need to add unit tests to these functions to check that works as expected. Otherwise we might break things without intention on refactors.

LGTM

brianmcgee:
@aldoborrero I agree. I have a wip branch where I'm making it easier to specify tests for nixos modules. I'll have a look at doing the same for apps this week.

PR #134, opened 2023-03-02
chore: move to nix-update and statix from nixpkgs-unstable

### LINKED ISSUE #133: statix/nix-update reference
Hello!
While preparing #121 I realize that we [use](https://github.com/nix-community/ethereum.nix/blob/main/nix/shell.nix#L12) `nix-update` and `statix` from stable nixpkgs input in devShell instead of their specific inputs. I found it because `nix-update` from stable channel doesn't support `--flake` argument.

I see two ways to resolve that:
* we use these packages from related inputs with `follows='nixpkgs-unstable'`, but we need to rebuild them cuz no caches 
* we use these packages from `nixpkgs-unstable` (and remove `nix-update` and `statix` inputs) because it tracks the last app versions very well, and we don't need to rebuild them locally

I'd prefer the second way. What do you think about that?

selfuryon:
My way to resolve #133

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,4 +1,5 @@
 if ! has nix_direnv_version || ! nix_direnv_version 2.2.0; then
     source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/2.2.0/direnvrc" "sha256-5EwyKnkJNQeXrRkYbwwRBcXbibosCJqyIUuz9Xq+LRc="
 fi
+nix_direnv_watch_file ./nix/shell.nix
 use flake
--- a/flake.nix
+++ b/flake.nix
@@ -39,14 +39,6 @@
     treefmt-nix = {
       url = "github:numtide/treefmt-nix";
     };
-    statix = {
-      url = "github:nerdypepper/statix";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
-    nix-update = {
-      url = "github:mic92/nix-update";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
   };
 
   outputs = inputs @ {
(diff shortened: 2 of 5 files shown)


.envrc
```diff
@@ -1,4 +1,5 @@
 if ! has nix_direnv_version || ! nix_direnv_version 2.2.0; then
     source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/2.2.0/direnvrc" "sha256-5EwyKnkJNQeXrRkYbwwRBcXbibosCJqyIUuz9Xq+LRc="
 fi
-use flake
\ No newline at end of file
+nix_direnv_watch_file ./nix/shell.nix
```
selfuryon, line 4: Found that direnv didn't track `shell.nix`

nix/checks.nix
```diff
@@ -1,22 +1,23 @@
...
+{
   self,
-  inputs,
+  inputs',
   lib,
   ...
-}: let
-  packagesModule = import ../packages args;
-in {
-  perSystem = psArgs @ {
```
selfuryon, line 9: We didn't use that, so I think we can remove them

flake.nix
```diff
@@ -39,14 +39,6 @@
...
       url = "github:numtide/treefmt-nix";
     };
-    statix = {
-      url = "github:nerdypepper/statix";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
-    nix-update = {
-      url = "github:mic92/nix-update";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
```
selfuryon, line 49: We don't use these inputs anyway, so we can remove them

aldoborrero (approved):
LGTM

brianmcgee (approved):
LGTM

PR #147, opened 2023-03-09
feat: add mkdocs and auto-document nixos module types

### LINKED ISSUE #50: How best to document everything?

### LINKED ISSUE #144: Document NixOS module options
[Disko](https://github.com/nix-community/disko) produces automated documentation for the NixOS options within their modules [like this](https://github.com/nix-community/disko/blob/master/doc.nix). 

I think we should follow a similar approach, but it will first require refactoring all of the types out of our modules so they can be isolated when running `lib.evalModules` otherwise you get complaints about options for networking and so on not being available.

### LINKED ISSUE #148: Docs: add an about section
We should provide some background on the project and what it is aiming to achieve. There will probably be some overlap with the README. Look at other nix-community projects for inspiration.

brianmcgee:
Introduces an initial mkdocs layout and auto documenting of NixOS modules options.

Also flattens the NixOS module directory structure as I don't feel it was giving us much and it was only serving to make import paths for shared code longer. 

Closes #50 
Closes #144 
Closes #148

- feat(modules): flatten directory structure
- feat(modules/geth): separate option types from the main module
- feat(modules/erigon): separate option types from the main module
- feat(modules/geth-bootnode): separate bootnode into separate module
- feat(modules/nethermind): separate options types from the main module
- feat(modules/prysm-beacon): separate options types from the main module
- fix(modules/geth-bootnode): adjust nat description so it does not conflict with doc generation
- feat(docs): generate docs for module options
- feat(docs): add initial mkdocs setup


brianmcgee:
FYI the build is failing due to a cachix bug it seems. @Mic92 has seen similar before and has reached out to @domenkozar with some info.

Issue #121, opened 2023-02-23
Add nix-update to workflows


selfuryon:
Manual updating is a little bit frustrating, `nix-update` is a huge step forward, but I suggest making a workflow similar to `Updates nix flakes` for updating packages (which I can update without setting specific version like in `web3signer`). 
Do PR in next several days

aldoborrero:
Speaking with @Mic92 about a possible way of achieving this with a Github Action would be:

1. We can take inspiration from  (nix-updater-action)[https://github.com/knl/niv-updater-action).
2. We can replace `niv` with `nix-update`.
3. We can list the packages in JSON with the command `nix flake show --json`
4. Iterate and update those we are interested in updating.
5. Done

brianmcgee:
@selfuryon have you had any time to look into this?

Issue #125, opened 2023-02-23
Package Teku from source (if it makes sense)


aldoborrero:
Same as #124 but for Teku.

aldoborrero:
@brianmcgee you have been battling lately with JVM and Nix, just evaluate if it's worthwhile or not. If it's not then close the issue please.

aldoborrero:
I'll mark this issue as closed for now. We can always get back to the idea if it makes sense.

Issue #124, opened 2023-02-23
Package Besu from sources (if it makes sense)


aldoborrero:
We're currently using a packaged version of Besu, but ideally, we should compile it from sources instead. 

Let's evaluate the difficulty of packaging Besu with Nix and determine if it's worthwhile not to do it (mostly, it will depend on how `gradle` is configured on their end and what tricks we may perform).

selfuryon:
I'll try to figure out how to do that (I haven't experienced with Java, but saw some packaged apps in nixpkgs)

aldoborrero:
@selfuryon I can help you with that if necessary.

selfuryon:
@aldoborrero Yeah, seems like I need some help with gradle plugins, how can I contact with you about that?

aldoborrero:
@selfuryon I'm available on Telegram or Matrix, whatever suits you better. I'll also include @brianmcgee as he recently packaged a Java app so his insights might be useful.

brianmcgee:
Relevant: https://bmcgee.ie/posts/2023/02/nix-what-are-fixed-output-derivations-and-why-use-them/

aldoborrero:
@brianmcgee you have been battling lately with JVM and Nix, just evaluate if it's worthwhile or not. If it's not then close the issue, please.

brianmcgee:
I'll have a look at some point this week

aldoborrero:
I'll mark this issue as closed for now. We can always get back to the idea if it makes sense.

PR #162, opened 2023-03-13
chore: update workflows

selfuryon:
Add `nix-update` as a workflow for updating packages.

Close: #121


.github/workflows/update-packages.yaml
```diff
@@ -0,0 +1,26 @@
...
+        uses: actions/checkout@v3
+      - name: Install Nix
+        uses: cachix/install-nix-action@v19
+        with:
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+      - name: Update packages
+        uses: selfuryon/nix-update-action@v0.1.0
+        with:
+          blacklist: "bls,blst,evmc,mcl,besu,teku,docs,foundry,mev-geth,web3signer"
```
selfuryon, line 20: Add to blacklist:
* `bls,blst,evmc,mcl` - they are libraries
* `besu,teku, web3signer` - they aren't built from source, so `nix-update` requires specifying version for them. Related to #124  #125 
* `mev-geth` doesn't follow semver, so `nix-update` can't update it.

Do we need to support mev-geth after merge?

.github/workflows/update-packages.yaml
```diff
@@ -0,0 +1,26 @@
+name: "Update Flake Packages ❄️"
```
selfuryon, line 1: Did a separate workflow because I think it's a good idea to have an opportunity to manually update only packages without inputs

.github/workflows/update-inputs.yaml
```diff
@@ -1,11 +1,10 @@
-name: "Update Nix Flakes ️❄️"
+name: "Update Flake Inputs ️❄️"
 on:
-  repository_dispatch:
```
selfuryon, line 3: Do we really need that or it's ok to delete that?

.github/workflows/update-packages.yaml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Install Nix
+        uses: cachix/install-nix-action@v19
+        with:
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+      - name: Update packages
+        uses: selfuryon/nix-update-action@v0.1.0
```
selfuryon, line 18: Use it as a first pre-release version. I also want to add some more features like gpg signs and so on. After that, I will update the version here.

brianmcgee (commented):
LGTM

brianmcgee (approved):
LGTM

PR #170, opened 2023-03-15
feat(ci-hercules): restrict ciSystems

selfuryon:
I checked `hercules-ci` and found out that we have long-running tasks because we haven't `aarch64-darwin` and `x86_64-darwin` agents to build them. I check the docs and found that we can restrict the [systems](https://flake.parts/options/hercules-ci-effects.html#opt-herculesCI.ciSystems) which we want to use for packages/checks build.
Tried to do that and seems it [works](https://hercules-ci.com/github/nix-community/ethereum.nix/jobs/923).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -36,6 +36,45 @@
         "type": "github"
       }
     },
+    "flake-parts_2": {
+      "inputs": {
+        "nixpkgs-lib": "nixpkgs-lib"
+      },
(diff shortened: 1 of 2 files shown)


brianmcgee:
This is quite handy until we can get a darwin CI runner.

brianmcgee (approved):
LGTM

PR #179, opened 2023-03-22
feat(modules): change backup provider from borg to restic

brianmcgee:
[Restic](https://restic.net/) is a better backup provider for a variety of different reasons, but the main one is that it supports S3 backends. Rather than support multiple providers I decided to just move from borg to restic directly. 

- feat(modules/backup): change backup provider from borg to restic
- feat(modules/restore): change backup provider from borg to restic
- docs: update backup and restore sections


selfuryon (approved):
LGTM, cool changes!
Not related to exactly this PR, but what do you think to move the [metadata logic/service](https://github.com/nix-community/ethereum.nix/blob/main/modules/backup/default.nix#L50) to geth/prysm/etc modules itself for reducing ["if" logic inside](https://github.com/nix-community/ethereum.nix/blob/main/modules/backup/default.nix#L75)?

brianmcgee:
> LGTM, cool changes! Not related to exactly this PR, but what do you think to move the [metadata logic/service](https://github.com/nix-community/ethereum.nix/blob/main/modules/backup/default.nix#L50) to geth/prysm/etc modules itself for reducing ["if" logic inside](https://github.com/nix-community/ethereum.nix/blob/main/modules/backup/default.nix#L75)?

I completely agree. I'm not happy with the current approach, I want to find a way of allowing each client module to specify it's own method of capturing metadata. 

Related to this I'm also considering creating a simple go program for capturing EL metadata and a separate one for capturing CL metadata as I want to also capture the network and chain id and so on which I think bash is not really suitable for.

PR #195, opened 2023-03-30
chore(mission-control): replace installToDevShell with a new config option

selfuryon:
Hello!
After update my system I found out that `mission-control` changed [the way how to work](https://github.com/Platonic-Systems/mission-control/pull/30/files) with devShell. Fix that before `Update Flake Inputs` task performs an update.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -240,11 +240,11 @@
     },
     "mission-control": {
       "locked": {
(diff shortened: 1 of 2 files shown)


aldoborrero (approved):
LGTM

PR #221, opened 2023-04-20
fix nixConfig in flake.nix

DavHau:
- It is unnecessary to put cache.nixos.org into the `extra-substituters` as it is already part of `substituters`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,11 +3,9 @@
 
   nixConfig = {
     extra-substituters = [
-      "https://cache.nixos.org"
       "https://nix-community.cachix.org"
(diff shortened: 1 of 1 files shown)


selfuryon:
@DavHau Can you sign your commit, pls? Without signed commits, we can't merge PRs

brianmcgee:
> * It is unnecessary to put cache.nixos.org into the `extra-substituters` as it is already part of `substituters`.

I've been copy pasting this around so many times now, don't remember where I originally got it from.

selfuryon (approved):

PR #230, opened 2023-04-22
Remove nonexistent ws.port value from Erigon allowedTCPPorts

RichardWarfield:
Currently the module for Erigon adds the value of the `ws.port` option to `allowedTCPPorts`, but this option does not exist either as an option in the module, nor on the Erigon command line.  Because the option does not exist, starting Erigon with `openFirewall = true` currently fails.  This PR removes the reference to `ws.port` in `allowedTCPPorts`.

### DIFF
--- a/modules/erigon/default.nix
+++ b/modules/erigon/default.nix
@@ -285,7 +285,6 @@ in {
               allowedTCPPorts =
                 [port authrpc.port torrent.port]
                 ++ (optionals http.enable [http.port])
-                ++ (optionals ws.enable [ws.port])
                 ++ (optionals metrics.enable [metrics.port]);
             }
         )


aldoborrero:
Hi, @RichardWarfield thanks for your contributions! In order to keep the flow going I think we can `cherry-pick` your commits and prepare separate PRs so we get those commits in the main branch so everybody can benefit us. 

@selfuryon can you help me with this?

RichardWarfield:
Please feel free.  Apologies, I've been travelling lately and just haven't
been able to attend to this. I actually didn't realize that pushing to the
branch would also enlarge the PR (oops).  I've added modules for some
additional components (prysm validator, mev boost) but please check whether
these are implemented the way you would like.  For example, I've added an
option to set the user and datadir for the Prysm services so the validator
and beacon clients can be set to the same user and access the same data
dir.  There may be a cleaner way to do this while continuing to use the
systemd state dir as the datadir but I'm not such an expert on systemd
options.

aldoborrero:
Closing this PR in favor of #280 and #279 (let's move the discussion there, if necessary)

PR #233, opened 2023-04-23
Fix erigon --ws

jotto:
Using websockets breaks because the module expects `ws.port` but doesn't define an arg for it... which is correct since Erigon doesn't have a `ws.port` flag.

### DIFF
--- a/modules/erigon/default.nix
+++ b/modules/erigon/default.nix
@@ -285,7 +285,7 @@ in {
               allowedTCPPorts =
                 [port authrpc.port torrent.port]
                 ++ (optionals http.enable [http.port])
-                ++ (optionals ws.enable [ws.port])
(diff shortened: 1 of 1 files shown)


aldoborrero (approved):
LGTM

aldoborrero:
On second thought, I think we can remove entirely the line: `++ (optionals ws.enable [ ])`.

As you correctly pointed out @jotto [everything runs under the RPC port](https://github.com/ledgerwatch/erigon#rpc-ports).

aldoborrero:
You're right, only removing `++ (optionals ws.enable [ ])` is fine, nothing else.

brianmcgee (approved):
Merging as it's a fix, we can clean up the empty array later

PR #234, opened 2023-04-23
Fix prysm build when nix env is go 1.2x

jotto:
* Prysm currently builds on go 1.19: https://github.com/prysmaticlabs/prysm/blob/08d6eccfb3bca915ff7d6369677b3f9ab9bacb55/.github/workflows/go.yml
* If using unstable or staging-next packages in Nix, you could end up with go 1.2x, which breaks Prysm due to this: https://github.com/trailofbits/go-mutexasserts/pull/2 (this fix is merged, but not yet in prysm)

### DIFF
--- a/packages/clients/consensus/prysm/default.nix
+++ b/packages/clients/consensus/prysm/default.nix
@@ -1,11 +1,11 @@
 {
   bls,
   blst,
-  buildGoModule,
+  buildGo119Module,
   fetchFromGitHub,
   libelf,
 }:
-buildGoModule rec {
+buildGo119Module rec {
   pname = "prysm";
   version = "4.0.1";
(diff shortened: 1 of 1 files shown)


selfuryon:
Currently, we are using stable version for prysm building where go is referenced to 1.19, but maybe it's a good thing to migrate to go 1.19 explicitly. @aldoborrero @brianmcgee What do you think about that?

P.S.: we need signed commits, we can't merge PR without that

brianmcgee:
It makes sense to pin to 1.19. @jotto in general we require signed commits but in this instance I'll go ahead and force merge this.

brianmcgee (approved):
LGTM

PR #242, opened 2023-04-24
chore: update packages daily

### LINKED ISSUE #239: Update frequency of workflow/update-packages.yaml to be executed daily
With the increasing amount of new packages and utilities we're adding to this repository, I think the best would be to, at least, run the workflow daily (that way we will cover any releases that are being done during midweek).

A separate conversation would be about how to proceed and how we will create a release policy like how nixpkgs does (i.e: stable, unstable).

aldoborrero:
Closes #239


selfuryon (approved):
LGTM

PR #250, opened 2023-04-26
chore(flake-parts): move to flake-parts.lib.mkFlake

selfuryon:
Migrated to `flake-parts.lib.mkFlake`, it's a default way to make flake, and it seems more clean for me.


brianmcgee (approved):
Yeah I don't know why we were doing it the other way. I've just been copy pasting this pattern everywhere.

PR #251, opened 2023-04-26
fix: sedge build

### LINKED ISSUE #201: sedge: update to v1.0.0
I can't update `sedge` to v1.0.0 because `go mod vendor` doesn't work correctly with it. I opened the issue about that https://github.com/NethermindEth/sedge/issues/272

selfuryon:
Fixed sedge build, now it works

Close #201

### DIFF
--- a/packages/utils/sedge/default.nix
+++ b/packages/utils/sedge/default.nix
@@ -3,11 +3,12 @@
   buildGoModule,
   fetchFromGitHub,
   mcl,
+  mockgen,
 }: let
   pname = "sedge";
   version = "1.1.0";
 in
-  buildGoModule rec {
+  buildGoModule {
     inherit pname version;
 
     src = fetchFromGitHub {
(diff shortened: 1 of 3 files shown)


packages/utils/sedge/default.nix
```diff
@@ -2,29 +2,41 @@
...
-    owner = "NethermindEth";
-    repo = pname;
-    rev = "v${version}";
-    sha256 = "sha256-U48ZM/J/Wj8chQvVB1DuqS3zl/qFMxDzlWS7Y3+U43Y=";
-  };
+    src = fetchFromGitHub {
+      owner = "NethermindEth";
+      repo = pname;
+      rev = "v${version}";
+      sha256 = "sha256-rIBHRiVZT3XXZB8zuFOoX02/g/DhBvIwa+sdis1XSYs=";
```
aldoborrero, line 18: @selfuryon let's change this to `sha` (we will update other packages as well)

packages/utils/sedge/default.nix
```diff
@@ -2,29 +2,41 @@
   bls,
   buildGoModule,
   fetchFromGitHub,
-  lib,
   mcl,
-}:
-buildGoModule rec {
+  mockgen,
+}: let
   pname = "sedge";
```
aldoborrero, line 8: @selfuryon by bringing these two variables, do you think nix-index will be able to update those two without issues?

aldoborrero (changes requested):
@selfuryon just only minor nitpicks but once clarified we can proceed to merge! Good work!

packages/utils/sedge/default.nix
selfuryon: Good catch, I didn't think about that, but I tested it and it works fine!

packages/utils/sedge/default.nix
selfuryon: done, changed it to `hash`

selfuryon:
Fixed a conflint

PR #257, opened 2023-04-27
feat: introduction of devshell

aldoborrero:
This PR introduces [`devshell`](https://github.com/numtide/devshell) rather than using `mission-control`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <div align="center" style="margin-top: 1em; margin-bottom: 3em;">
-  <h1>ethereum.nix = Ethereum + Nix 🫶</h1>
+  <h1>ethereum.nix = Ethereum 🫶 Nix</h1>
 </div>
 
 <p align="center">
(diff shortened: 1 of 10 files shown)


selfuryon (dismissed):
LGTM, but just curious, why not devenv or similar projects?

aldoborrero:
@selfuryon I chose `devshell` as it's quite similar to `mission-control` but removes extra mambo jambo that regular `mkShell` introduces into the system variables when loading the dev environment.

And why not `devenv`? Mostly it's a personal preference, I do believe it introduces too much that, probably, we won't need in the near term. We don't need features like processes, languages, containers, etc. Even if we do someday, I prefer to incorporate more fine-grained tools like [`process-compose`](https://github.com/F1bonacc1/process-compose) or another similar tooling. I prefer to favor the Unix philosophy (do one thing very well).

Also with `devshell` I can optimize the repository much faster if necessary as I'm working with Numtide (the owner of the repository) these days.

As with everything, it's not set in stone, so we can evaluate at a later date another tool that suits us better (even `devenv`, I'm not excluding it by any means). I'm always open to exploring new ideas.

brianmcgee:
@aldoborrero can you sign the commits?

aldoborrero:
Mmm, it seems `hercules-ci` is behaving weirdly today...

selfuryon (approved):
LGTM

PR #259, opened 2023-04-28
chore: ignore repl-results

aldoborrero:
Just ignore `repl-result` when using `nix repl` and `:bl <derivation>` command.


selfuryon (approved):
LGTM

PR #271, opened 2023-05-08
chore: removal of 41North copyright

aldoborrero:
Removes deprecated copyright notices of 41North and moves `option-docs.nix` into `mkdocs.nix`.


brianmcgee (approved):
LGTM

PR #279, opened 2023-05-11
Add service for Prysm validator

aldoborrero:
Take commits from PR #230


aldoborrero:
Will introduce tests as well for this module

PR #280, opened 2023-05-11
Add mev-boost module

aldoborrero:
Introduces mev-boost nixos module

### DIFF
--- a/modules/mev-boost/default.test.nix
+++ b/modules/mev-boost/default.test.nix
@@ -1,0 +1,28 @@
+{
+  systems = ["x86_64-linux"];
+
+  module = {
+    name = "mev-boost";
+
+    nodes = {
+      basic = {
+        virtualisation.cores = 2;
+        virtualisation.memorySize = 4096;
+
+        services.ethereum.mev-boost.test = {
+          enable = true;
+          args = {
+            addr = "localhost:18550";
+            network = "goerli";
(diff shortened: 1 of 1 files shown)


brianmcgee (changes requested):
Can you add a quick startup test like with Geth, it should help to avoid obvious broken packages.

aldoborrero:
Will do

selfuryon:
@aldoborrero What do you think to use `freeformType` ([link](https://nixos.org/manual/nixos/stable/#sec-freeform-modules)) for args (and maybe something else) in all modules? It allows using undeclared parameters too. I found one example: nix daemon module ([usage](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/services/misc/nix-daemon.nix#L473), [declaration](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/services/misc/nix-daemon.nix#L95))

aldoborrero:
@selfuryon nice discovery, let's give it a try! 

Let's make sure though we beef up the tests and also if we can simplify some logic in the process.

aldoborrero:
I'll add the `freeformType` support in another PR as it will touch multiple files.

selfuryon (approved):
LGTM

brianmcgee (approved):
LGTM

PR #311, opened 2023-06-12
chore(erigon): update erigon v.2.45.1 -> v2.45.2

brianmcgee:


selfuryon:
It's strange that there are no changes in hash here

brianmcgee:
The vendor hash will only change if one of the module dependencies changes. I guess they've only been making adjustments to the erigon source.

brianmcgee:
I needed to manually break the source hash and the vendor hash to get a hash mismatch to then update with the correct hashes. 

I don't understand why we weren't getting a mismatch before after changing the version.

selfuryon (approved):
LGTM

brianmcgee:
@selfuryon after speaking with @NinjaTrappeur:

> The fixed output derivations (FOD, the derivation for which you give an expected sha256) are output-adressed. Their hash is not computed from their inputs but via the specified sha256.
>
> Here, you did not change the output hash, Nix considered the output hash as derivation hash. You already had this path in your store, Nix did not try to rebuild it, you did not see the hash mismatch.
>
> You do not face this issue with "regular" derivations because changing any input changes the derivation output hash.

So moral of the story is always break the hash on update to get a mismatch and hence figure out what the new hash should be.

PR #337, opened 2023-07-07
feat: add slither

aldoborrero:
Introduces [Slither](https://github.com/crytic/slither). Not the latest version as we need to update a couple of libraries, but closer enough.


brianmcgee (approved):
LGTM

PR #340, opened 2023-07-10
feat(slither): update slither and add a crytic-compile as a lib

selfuryon:
We couldn't update `slither` because unstable nixpkgs has a `0.3.0` crytic-compile version, but the latest `slither` needs at least `0.3.3`. I added `crytic-compile` as a package dependency until nixpkgs will merge [this PR](https://github.com/NixOS/nixpkgs/pull/230067)

### DIFF
--- a/packages/default.nix
+++ b/packages/default.nix
@@ -9,13 +9,14 @@
 
   perSystem = {
     self',
+    inputs',
     pkgs,
     system,
     ...
   }: let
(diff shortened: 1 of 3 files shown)


aldoborrero (approved):
LGTM

I was in the process of preparing a PR with the fixes, but discarding mine seeing it's already on `nixpkgs` :)

PR #348, opened 2023-07-25
chore: update nimbus 23.6.1 -> 23.7.0

selfuryon:
I also faced with some problems during nimbus upgrade to a new version, i'll try to investigate that in a separate branch


selfuryon:
Found out that it maybe some bugs in nim from comments [here](https://github.com/status-im/nim-ssz-serialization/blob/master/ssz_serialization/merkleization.nim#L201). After switching to the latest nim version, it builds successfully.

brianmcgee (approved):
LGTM

PR #356, opened 2023-08-16
feat: add eth2-val-tools

selfuryon:
Current [holesky genesis onboarding guide](https://notes.ethereum.org/@ethpandaops/Holesky-onboarding) mentions [eth2-val-tools](https://github.com/protolambda/eth2-val-tools) very often, so I decided to add it too


brianmcgee (approved):
LGTM

PR #370, opened 2023-09-10
chore: erigon: 2.48.1 -> 2.49.0

selfuryon:
Currently, it can't be built, need to investigate the reason, but maybe this is related: https://github.com/mattn/go-sqlite3/issues/1165

PR #345, opened 2023-07-20
nethermind: 1.19.3 -> 1.25.3

selfuryon:
I tried to upgrade nethermind, but I stuck with a problem with [PackageSourceMapping](https://github.com/NethermindEth/nethermind/commit/929833530bb1b48d68d7f49c33475529677fbf3d) in a new release.

PackageSourceMapping was enabled to use two sources for nuget packages:
* `nuget.org` as a default option
* `nugettest.org` for `Nethermind.DotNetty` packages

Problems:
* `nix run .#nethermind.fetch-nethermind-deps packages/clients/execution/nethermind/nuget-deps.nix` works, but produce an empty hash for packages from `nugettest.org`. Looks like an embedded script doesn't work with `nuget.config`
* if `PackageSourceMapping` is enabled, I can't restore packages in `buildDotnetModule` build phase anymore with my own source: internal `env dotnet restore src/Nethermind/Nethermind.sln ... --source <nix-store>/nuget-source/lib` ignores source option `PackageSourceMapping is enabled, the following source(s) were not considered <nix-store>/nuget-source/lib`
* if I disable `PackageSourceMapping`, dotnet can't find any `nugettest.org` related packages...

I didn't find an option for a cache directory with `PackageSourceMapping` configuration, but maybe I just missed it (I've never worked before with dotnet). Trying to fix...

PR #371, opened 2023-09-10
chore(workflows/update-packages): blacklist some packages

selfuryon:
Consider to disable auto update for these packages:
* mev-boost: until they will make a new release tag. Actually, it's very strange that `nix-update` try to update it to a pre-release version (1.6.0 -> 1.6.4844-dev2)
* erigon: until we fix its build #370 [Done]
* nethermind: until we fix build with Package Source Mapping #345


selfuryon:
I fixed `erigon` in #370 and re-enabled the after-update for it. Merge after #370

brianmcgee (approved):
Lgtm

PR #376, opened 2023-09-14
Flake organization improvements

aldoborrero:
I plan to resume more work related to Ethereum.nix on Monday 25th, but for now, I just wanted to simplify the overall design of the flake.

We will incorporate more changes soon to take advantage of flake-parts for other stuff.


selfuryon (approved):
LGTM
What do you think about using [haumea](https://github.com/nix-community/haumea) like a replacement for rakeLeaves?

aldoborrero:
@selfuryon I was not aware of haumea, I'll give it a try in a separate branch (thanks for pointing it out)

PR #393, opened 2023-10-02
Fix #392

### LINKED ISSUE #392: `findEnabled` can cause infinite recursion
### Description
[findEnabled](https://github.com/nix-community/ethereum.nix/blob/433624d8bca7bf83b723a2ec716178770a9506f6/modules/lib.nix#L102) will attempt to recurse through all attributes in the provided tree until it finds an attribute set that contains `{ enabled = true; }` in each of the branches. This works in the "normal" case when it it used on a configuration like:

```nix
services.ethereum.geth.sepolia = {
    enable = true;
    ...
  };

  services.ethereum.geth.goerli = {
    enable = true;
    ...
  };
```

However if you disable any of the "top level" services, things start to break down. Since all services contain a `package` option, `findEnabled` will attempt to recurse into the derivation itself which will result in an infinite recursion as derivations always contain a reference to themselves (`out`). Additionally, since `findEnabled` is now searching deeper than it was actually intended for, it may now come across submodules that are enabled (such as `geth.<name>.args.http.enable` or even `geth.<name>.backup.enable`) despite the fact that they won't actually do anything since the outer module is disabled. This can result in additional evaluation errors as bad data is being fed into the backup/restore modules.

### Steps to reproduce

1. Attempt to evaluate a module containing:
```nix
services.ethereum.geth.sepolia = {
    enable = false;
    backup = {
      enable = false; # not respected since findEnabled gets confused by get service being disabled
      restic.passwordFile = ""; # needed since backup/restore modules think backup/restore is enabled
    };
    restore = {
      enable = false; # not respected since findEnabled gets confused by get service being disabled
      restic.passwordFile = ""; # needed since backup/restore modules think backup/restore is enabled
      snapshot = ""; # needed since backup/restore modules think backup/restore is enabled
    };
};
```
2. Observe infinite recursion

### Proposed solution

Modify `findEnabled` so that it only recurses if the current attribute set does not contain the `enabled` attribute regardless of its value. We still only add to `sum` if `enabled == true`, however for the case when `enabled` is present but `false`, we will consider that branch done and stop recursing. This will avoid us searching deeper than was intended which resulted in the bugs highlighted above.

scottbot95:
Implements my proposal from #392

### DIFF
--- a/modules/lib.nix
+++ b/modules/lib.nix
@@ -106,7 +106,7 @@ in {
       in
         if (attrByPath ["enable"] false val)
         then (sum // {"${pathStr}" = val;})
-        else if (isAttrs val)
+        else if ((isAttrs val) && !(val ? enable))
         then (recurse sum path val)
         else sum;
 


aldoborrero (approved):
LGTM. Maybe as I said on #392 we could consider adding a check to `isDerivation` and avoid recursing in such instances as well.

scottbot95:
I had that at one point, but checking for the presence of `enabled` should prevent us from seeing a derivation in the fist place. I guess there's no reason not to include the extra checks though.

Ideally we would probably do whatever NixOS modules do to detect infinite recursion, but I can't be bothered to figure that out at this time lol.

aldoborrero:
Hey it works for now. We can improve the code later if necessary 😉

PR #400, opened 2023-10-21
Add tests for Lighthouse Validator

scottbot95:
Add basic test for Lighthouse Validator. Pretty much just tests that the service starts up and doesn't crash immediately (see [my comment on my last PR](https://github.com/nix-community/ethereum.nix/pull/394#issuecomment-1771328926))


scottbot95:
Not sure why mev-boost failed the build... I can't see anything in the logs available to me and it builds just fine locally.

aldoborrero:
@scottbot95 see the comment I've written and [let's decide on a better approach](https://github.com/nix-community/ethereum.nix/pull/394#issuecomment-1778987567) rather than starting and stopping twice the service (which I'm totally fine if we introduce more checks in the middle).

scottbot95:
@aldoborrero I updated the test for lighthouse validator tests per your suggestions. It's probably worth creating an issue to follow-up and update other tests/fill-in missing tests with this pattern. I wonder if it's also worth trying to formalize it in a `CONTRIBUTING.md` doc? Or maybe writing a re-usable python module or something? I'm not very familiar with python so not sure how feasible the latter is.

aldoborrero:
@scottbot95 I'm going to close this PR as I never merged it (as I was unsure of adding timeouts). Let me know if we should review it appropiately.

PR #417, opened 2023-11-27
fix: override go modules for eigenlayer-cli

selfuryon:
Turns out, that I can't build eigenlayer-cli if I remove `vendorhash`, because vendored modules require `mockgen` and `go generate` for building and passing the tests. So I did some fancy override for making the build to work.

Another case related to building depends with `buildbot` and `devour-flake`?


aldoborrero:
Now, it should be solved. I'm discovering that `eigenlayer` doesn't compile well for `aarch64-darwin` as [we can see here](https://buildbot.nix-community.org/#/builders/335/builds/12).

aldoborrero:
@selfuryon I also discovered by looking at the [buildbot logs](https://buildbot.nix-community.org/#/builders/423/builds/2) that on `darwin` it doesn't compile either. We can just leave it as it is with only `x86_64-linux` and track the issue separately. Let me know.

selfuryon:
I did that in #406

PR #419, opened 2023-11-29
fix: darwin libusb issues with foundry

aldoborrero:
This solves the upstream issue of foundry not finding libusb on darwing architectures by just commenting the platforms.

### DIFF
--- a/packages/default.nix
+++ b/packages/default.nix
@@ -61,7 +61,9 @@
 
       # Dev
       foundry = inputs.foundry-nix.defaultPackage.${system}.overrideAttrs (_oldAttrs: {
-        meta.platforms = [system];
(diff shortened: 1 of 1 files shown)


selfuryon:
I fixed that in upstream: https://github.com/shazow/foundry.nix/pull/29 and https://github.com/shazow/foundry.nix/pull/28. Looks like we need to ask maintainer to run again monthly workflow :) I will do

aldoborrero:
Great, but also I've discovered that `libusb` is not included in the darwin platform as I pointed on [this issue](https://github.com/shazow/foundry.nix/issues/23#issuecomment-1832032254) (see the [build logs here](https://buildbot.nix-community.org/#/builders/336/builds/18/steps/1/logs/stdio))

selfuryon:
I did that in #406

PR #420, opened 2023-12-10
fix: import `poetry2nix` properly

lourkeur:
I think overlays are the intended way to consume poetry2nix. [flake.parts recommends consuming an overlay at the `pkgs` point.](https://flake.parts/options/flake-parts-easyoverlay)


selfuryon:
Thanks for your PR! I moved your changes in #406 (can't pass the tests without that).

PR #423, opened 2023-12-29
nimbus: Also build nimbus_validator_client

catwith1hat:
This commit make the nimbus package build the stand-alone validator client by default. The stand-alone validator is documented here: https://nimbus.guide/validator-client.html.

### DIFF
--- a/packages/default.nix
+++ b/packages/default.nix
@@ -97,7 +97,10 @@
       lighthouse.bin = "lighthouse";
 
       # consensus / nimbus
-      nimbus.bin = "nimbus_beacon_node";
+      nimbus = {
+        nimbus-beacon-node.bin = "nimbus_beacon_node";
+        nimbus-validator-client.bin = "nimbus_validator_client";
+      };
 
       # execution clients
       besu.bin = "besu";
--- a/packages/clients/consensus/nimbus/default.nix
+++ b/packages/clients/consensus/nimbus/default.nix
@@ -7,7 +7,7 @@
   cmake,
   pkgs,
   writeShellScriptBin,
-  buildFlags ? ["nimbus_beacon_node"],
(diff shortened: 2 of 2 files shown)


selfuryon:
Hello! Thanks for the PR!! I see that there are two separate binaries, can you add applications for them too, please? like [here](https://github.com/nix-community/ethereum.nix/blob/main/packages/default.nix#L86)

catwith1hat:
I added the two applications to the list. Is it okay that I removed `nimbus.bin` in favor of `nimbus.nimbus_beacon_chain.bin`? I am not using this mechanism, so not sure what's the proper way.

selfuryon:
> I added the two applications to the list. Is it okay that I removed `nimbus.bin` in favor of `nimbus.nimbus_beacon_chain.bin`? I am not using this mechanism, so not sure what's the proper way.

Yeah, that's ok!

packages/default.nix
```diff
@@ -97,7 +97,10 @@
       lighthouse.bin = "lighthouse";
 
       # consensus / nimbus
-      nimbus.bin = "nimbus_beacon_node";
+      nimbus = {
+        nimbus_beacon_node.bin = "nimbus_beacon_node";
```
selfuryon, line 101:
```suggestion
        nimbus-beacon-node.bin = "nimbus_beacon_node";
```

packages/default.nix
```diff
@@ -97,7 +97,10 @@
...
+        nimbus_validator_client.bin = "nimbus_validator_client";
```
selfuryon, line 102:
```suggestion
        nimbus-validator-client.bin = "nimbus_validator_client";
```

selfuryon (commented):
Just small fix

selfuryon (approved):
LGTM!

PR #425, opened 2024-01-07
besu: 23.10.1 -> 23.10.3-hotfix

catwith1hat:
Besu had a global outage at block 18947893. This fixes the issue.


selfuryon (approved):
LGTM

PR #429, opened 2024-01-23
flake: compress dependency tree

zimbatm:
Avoid duplicate inputs in the flake.lock.

As a consumer of the flake, it makes it easier to pass my own version
of flake-utils, systems, ...


brianmcgee (approved):
lgtm

PR #434, opened 2024-01-25
feat: implement Nimbus beacon service

iofq:
Implements Nimbus Beacon service (validator service to follow).

Two things I think could be contentious:

* Using `ExecPreStart` for the checkpoint sync. Nimbus has this implemented somewhat awkwardly IMO, in that it's a separate command. `ExecPreStart` has a timeout of 15 mins; in all my testing the checkpoint finished downloading in 2 or 3 mins, but had it not completed before the timeout, the systemd service will fail to start. We can as well adjust that 15 minute timeout to be longer or infinite.

* Opening REST and Metrics ports by default. This seems to be done by default in Lighthouse at least, which i used for reference for a lot of this, however I would argue that the majority of users would not want these exposed and as such maybe it should be a more explicit enable, either within the module or done outside of it. Not sure what the best UX is here.

### DIFF
--- a/docs/nixos/running-nimbus-beacon.md
+++ b/docs/nixos/running-nimbus-beacon.md
@@ -1,0 +1,86 @@
+# Running Nimbus Beacon
+
+One or more [Nimbus Beacon](https://nimbus.guide) services can be configured with the `services.ethereum.nimbus-beacon` prefix.
+
+```nix title="server.nix"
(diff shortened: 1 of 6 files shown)


iofq:
rebased and signed commit

brianmcgee (approved):
lgtm

PR #439, opened 2024-01-25
fix: restore flake show

zimbatm:
poetry2nix reads the pyproject.toml. but that creates an IFD. by moving
the fetch to eval time, we avoid that issue.


brianmcgee (approved):
Lgtm

Issue #258, opened 2023-04-28
Introduce flake dev pattern


aldoborrero:
Right now if we run `nix flake metadata`, we obtain the following output:

<details>
<summary>Terminal outputs of Inputs</summary>

```bash
thinkpad-x1 :: Dev/nix-community/ethereum.nix2 ‹feat/devshell*› % nix flake metadata
Resolved URL:  git+file:///home/aldo/Dev/nix-community/ethereum.nix2
Locked URL:    git+file:///home/aldo/Dev/nix-community/ethereum.nix2?ref=refs%2fheads%2ffeat%2fdevshell&rev=21d7880e1e8c95b75209115c7b86c32c7985b714
Description:   A reproducible Nix package set for Ethereum clients and utilities
Path:          /nix/store/q2cn8xv7d4wvv1nrg16ngj57a6ai4jpn-source
Revision:      21d7880e1e8c95b75209115c7b86c32c7985b714
Revisions:     166
Last modified: 2023-04-28 10:27:16
Inputs:
├───devshell: github:numtide/devshell/3e0e60ab37cd0bf7ab59888f5c32499d851edb47
│   ├───flake-utils: github:numtide/flake-utils/846b2ae0fc4cc943637d3d1def4454213e203cba
│   └───nixpkgs follows input 'nixpkgs'
├───flake-compat: github:edolstra/flake-compat/35bb57c0c8d8b62bbfd284272c928ceb64ddbde9
├───flake-parts: github:hercules-ci/flake-parts/dcc36e45d054d7bb554c9cdab69093debd91a0b5
│   └───nixpkgs-lib follows input 'nixpkgs'
├───flake-root: github:srid/flake-root/f1c0b93d05bdbea6c011136ba1a135c80c5b326c
├───foundry-nix: github:shazow/foundry.nix/e3e0951e0cb668e7796f1fde2926ff32dc544d49
│   ├───flake-utils: github:numtide/flake-utils/3cecb5b042f7f209c56ffd8371b2711a290ec797
│   └───nixpkgs follows input 'nixpkgs'
├───hercules-ci-effects: github:hercules-ci/hercules-ci-effects/15ff4f63e5f28070391a5b09a82f6d5c6cc5c9d0
│   ├───flake-parts: github:hercules-ci/flake-parts/c13d60b89adea3dc20704c045ec4d50dd964d447
[...]

PR #441, opened 2024-01-29
flake: remove dependency on mynixpkgs

zimbatm:
I ported a bunch of the packages over to keep this project's flake.lock
reasonable in size.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -133,28 +133,6 @@
         "type": "github"
       }
     },
-    "haumea": {
-      "inputs": {
-        "nixpkgs": [
-          "mynixpkgs",
-          "nixpkgs"
-        ]
-      },
-      "locked": {
-        "lastModified": 1685133229,
-        "narHash": "sha256-FePm/Gi9PBSNwiDFq3N+DWdfxFq0UKsVVTJS3cQPn94=",
-        "owner": "nix-community",
-        "repo": "haumea",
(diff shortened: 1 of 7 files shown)


aldoborrero:
@zimbatm how do you determine reasonably the size and the impact of each input in a flake? I know the mantra of the less the better in terms of the evaluation but is there any tool to properly measure the impact?

For this particular case, I don't like to introduce those packages that don't belong strictly to Ethereum.nix, said that there are other quick wins like removing `flake-root` input and we can continue the work I've made on the issue #258

zimbatm:
I don't think there is necessarily a correct answer; it depends on how you weigh things.

What I noticed is that ethereum.nix pulls all of mynixpkgs but only uses a very small subset of it. Also mynixpkgs only has a single maintainer. That means that coordinating package updates between both repos creates more overhead for maintenance. In that case, my preference is to vendor things. Especially if the idea is that those packages will be in nixpkgs eventually.

aldoborrero:
@zimbatm let's put effort into upstreaming those packages directly to `nixpkgs` instead if you don't mind. I think they are valuable as they are related to `mkdocs` and are useful plugins.

zimbatm:
https://github.com/NixOS/nixpkgs/pull/285561

aldoborrero:
@zimbatm I'm closing this PR as we will have soon the new packages available on `nixpkgs`.

PR #442, opened 2024-01-29
flake: remove easyOverlay

zimbatm:
Work around issues caused by https://github.com/NixOS/nixpkgs/pull/269551


zimbatm:
This still needs some work, but I wanted feedback on the direction. Is it ok to remove support for overlays?

aldoborrero:
I prefer #443

PR #443, opened 2024-01-29
fix: remove easyOverlay

brianmcgee:
Create the default overlay manually as https://github.com/NixOS/nixpkgs/pull/269551 seems to have caused some issues with easyOverlay where it's trying to touch an i686 package when that system is not in the listed systems.


aldoborrero:
I prefer this solution contrary to removing it

PR #450, opened 2024-02-01
ci(Mergify): configuration update

aldoborrero:
This change has been made by @aldoborrero from Mergify config editor.


brianmcgee (approved):
Lgtm

PR #453, opened 2024-02-04
fix: nimbus-beacon rest.allow-origin bad default

iofq:
If `rest.enable` is true and `rest.allow-origin` is not set, `allow-origin` defaults to null which can't be coerced to a string when building CLI flags for Nimbus. This MR fixes this behavior by adding a null check before building the flag.


iofq:
Found and added commit for second bug, where `data-dir` flag was duplicated if it was non-default.

brianmcgee (approved):

PR #454, opened 2024-02-04
Add ReadWritePaths for datadir

tthebst:
I had issues running lighthouse with its datadir on a separate ssd `/mnt/extra/`. Adding `ReadWritePaths` solved it but I only verified it for lighthouse but might be useful for the others as well.

### DIFF
--- a/modules/erigon/default.nix
+++ b/modules/erigon/default.nix
@@ -103,6 +103,7 @@ in {
                 {
                   User = serviceName;
                   StateDirectory = serviceName;
+                  ReadWritePaths = [ cfg.args.datadir ];
                   ExecStartPre = mkIf cfg.subVolume (mkBefore [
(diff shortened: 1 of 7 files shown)


aldoborrero:
@tthebst the `treefmt` check has failed, would you mind running it? :)

aldoborrero:
Hi @tthebst make sure to compact those two commits into one and add a meaningful description to the commit (like "feat: add support to ReadWritePaths for datadir" or something like that).

Also, review the tests for lighthouse as those are failing.

tthebst:
Yes I saw that some tests fail so I put it back into draft.

aldoborrero:
@tthebst does this still apply? I'm going to close the PR as it has been more than 11 months now in draft. Let me know if we should re-open it.

Issue #126, opened 2023-02-23
Add support for foundry


aldoborrero:
Adding support for [foundry](https://github.com/foundry-rs/foundry/) would be very nice! Especially if we package it adequately from sources.

I know [`foundry.nix`](https://github.com/shazow/foundry.nix) effort exists; maybe we should talk with @shazow and try to merge it here?

shazow:
Not opposed to having it merged here, but a few caveats/warnings:

First, foundry.nix is packaging from binaries rather than from source. Last time I tried to package from source, there were too many untagged dependencies that it was hard to make it build properly in an immutable way (namely ethers-rs and some other stuff). I haven't checked recently what the state of this is now, I hope it's better now?

Second, there's still no tagged version releases for foundry, only nightly commit binary releases, which are ephemeral (get cleaned up after some number of weeks).

To work around this, foundry.nix does nightly autoupdating to pull the latest nightly release. In addition, we negotiated adding a monthly non-ephemeral nightly release (which unfortunately has been flaky), so we have a `monthly` branch that points to that one.

Not sure if it would be appealing for you to pull in all of these moving pieces. :) But you're very welcome to pull in any subset. I'm happy to generally migrate to this repo if we can achieve somekind of parity here, or keep both running in parallel until we can.

Ideally I was waiting for when there are tagged releases across the dependency set before I was going to try again to repackage.

shazow:
Sounds good!

If you want releases that aren't going to break in a few weeks, I suggest using the `monthly` branch for now. :) (`github:shazow/foundry.nix/monthly`)

PR #457, opened 2024-02-06
foundry: init at nightly-0688b5a

mitchmindtree:
This introduces a new built-from-source `foundry` package and renames the old binary package from the `foundry.nix` repo to `foundry-bin` following the `nixpkgs` convention.

### svm-rs

One quirk of `foundry` builds is the `svm-rs` crate dependency, which by default fetches a list of all available `solc` binaries for each platform`build.rs`. As there is no hash or versioning of the file itself, the file can change between fetches (e.g. when new solc versions are added).

Fortunately svm-rs exposes a `SVM_RELEASES_LIST_JSON` env var which allows for pointing to a local file path instead making support for Nix feasible without having the package break each time a new `solc` version is released. A script called `update-svm-lists.sh` is included that allows for downloading the latest lists for linux and darwin.

also cc @shazow you might be interested!

### DIFF
--- a/.github/workflows/update-packages.yaml
+++ b/.github/workflows/update-packages.yaml
@@ -17,7 +17,7 @@ jobs:
         with:
           # TODO: remove nethermind after we fix build for them
           # TODO: remove mev-boost after they make a new release tag
-          blacklist: "staking-deposit-cli,dreamboat,bls,blst,evmc,mcl,besu,teku,docs,foundry,web3signer,mev-boost-prysm,vscode-plugin-consensys-vscode-solidity-visual-editor,vscode-plugin-ackee-blockchain-solidity-tools,mev-boost,nethermind"
+          blacklist: "staking-deposit-cli,dreamboat,bls,blst,evmc,mcl,besu,teku,docs,foundry-bin,web3signer,mev-boost-prysm,vscode-plugin-consensys-vscode-solidity-visual-editor,vscode-plugin-ackee-blockchain-solidity-tools,mev-boost,nethermind"
           sign-commits: true
           gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
           gpg-passphrase: ${{ secrets.GPG_PASSPHRASE }}
--- a/packages/default.nix
+++ b/packages/default.nix
@@ -71,7 +71,8 @@
       zcli = callPackage ./utils/zcli {};
 
       # Dev
-      foundry = inputs.foundry-nix.defaultPackage.${system}.overrideAttrs (_oldAttrs: {
+      foundry = callPackageUnstable ./dev/foundry {};
+      foundry-bin = inputs.foundry-nix.defaultPackage.${system}.overrideAttrs (_oldAttrs: {
         # TODO: Uncomment when https://github.com/shazow/foundry.nix/issues/23
         # meta.platforms = [system];
         meta.platforms = ["x86_64-linux" "aarch64-linux"];
--- a/packages/dev/foundry/default.nix
+++ b/packages/dev/foundry/default.nix
@@ -1,0 +1,81 @@
+{
+  lib,
+  stdenv,
+  darwin,
+  fetchFromGitHub,
+  installShellFiles,
+  libusb1,
+  pkg-config,
+  rustPlatform,
+}:
+rustPlatform.buildRustPackage rec {
+  pname = "foundry";
+  version = "nightly-${builtins.substring 0 7 src.rev}";
+
+  src = fetchFromGitHub {
+    owner = "foundry-rs";
+    repo = "foundry";
+    rev = "0688b5ad19a637303c038d1a66aec62a73713e20";
+    hash = "sha256-OIsUzJVNcb2nVCYU/BdGGGICEg9Cr9LXc8zzN2JSb8g=";
+  };
+
+  cargoLock = {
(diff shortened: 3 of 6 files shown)


aldoborrero:
Hi @mitchmindtree thanks for the PR!  Awesome work!

In the past @shazow and I, discussed adding support directly to Foundry here in #126 (just in case to give you more context).

In the upcoming weeks, I'm about to work on improving the existing automation for auto-updating the packages without manual intervention (the existing solution works partially and there are a lot of packages in the blacklist these days). 

Foundry is a special beast with no stable releases and flaky monthly ones but in the context of Ethereum.nix I think ideally we should support only monthly ones (and treat it like the stable release). I would like to hear your opinions (@shazow and @mitchmindtree) on this specific subject (for now we will focus on updating it manually whenever it makes sense).

In terms of the PR itself, I would also add the removal of `foundry.nix` input.

shazow:
That's awesome.

I'm a big +1 to adding a build-from-source derivation for foundry, this would allow us to maintain stable pins regardless of the unreliable official release assets. Bonus points if we can cache the builds somehow (cachix?).

If ethereum.nix doesn't want to merge/maintain it at this time, I'm happy to include this in https://github.com/shazow/foundry.nix until it's sufficiently gestated. :)

Also, as mentioned in #126, I'm also generally happy to merge my foundry.nix into ethereum.nix and hand over the reins/collaborate here moving forward.

Whatever works best. :)

mitchmindtree:
> In the upcoming weeks, I'm about to work on improving the existing automation for auto-updating the packages without manual intervention

@aldoborrero Awesome! I was planning on checking in about this :heart: 

> Foundry is a special beast with no stable releases and flaky monthly ones but in the context of Ethereum.nix I think ideally we should support only monthly ones (and treat it like the stable release).

I empathise with the concern around foundry having no versioned, stable releases... that said, I'm not sure how much benefit only selecting a new nightly once a month would have over selecting a new one at whatever ethereum.nix's regular update schedule is? :thinking: 

I think as long as the auto-update process at least tries to build the updated packages (and do the check phase) before committing them, that's totally fine. I imagine we simply skip the nightly versions that fail to build?

If there are obscure runtime issues due to foundry instability, I guess there's not much we can do about that whether we update monthly or nightly... at least if we update more regularly, ethereum.nix users have more options of different versions (by using ethereum.nix at different commits).

In general, I imagine trying to decide on release schedules for downstream users might be a tricky line to walk as different folks might have different ideas of what a good schedule is for different pkgs? Perhaps ethereum.nix might be best just making available whatever it can from upstream at whatever rate it's comfortable with, and then let downstream decide what versions of things they want to pin to by choosing their own commits / when to run `nix flake update`? Kinda like how most people seemingly tend to use `nixos-unstable` to get the latest patches/updates/features, and just update the input when they're ready to?

> Bonus points if we can cache the builds somehow (cachix?).

@shazow I *think* this is already happening using the `nix-community` cache (the substituter that gets added when you build the `flake`) though I could be wrong, I haven't looked too closely! I originally opened PRs at nixpkgs so we could get this and `reth` in the default nixos cache, but neither have had any action yet :sweat_smile: We can probably move a bit quicker here anyways?

aldoborrero:
> I empathise with the concern around foundry having no versioned, stable releases... that said, I'm not sure how much benefit only selecting a new nightly once a month would have over selecting a new one at whatever ethereum.nix's regular update schedule is? 🤔
> 
> I think as long as the auto-update process at least tries to build the updated packages (and do the check phase) before committing them, that's totally fine. I imagine we simply skip the nightly versions that fail to build?
> 
> If there are obscure runtime issues due to foundry instability, I guess there's not much we can do about that whether we update monthly or nightly... at least if we update more regularly, ethereum.nix users have more options of different versions (by using ethereum.nix at different commits).

Fair enough. Let's then try to keep the package updated every day and skip those that fail to build.

> In general, I imagine trying to decide on release schedules for downstream users might be a tricky line to walk as different folks might have different ideas of what a good schedule is for different pkgs? Perhaps ethereum.nix might be best just making available whatever it can from upstream at whatever rate it's comfortable with, and then let downstream decide what versions of things they want to pin to by choosing their own commits / when to run `nix flake update`? Kinda like how most people seemingly tend to use `nixos-unstable` to get the latest patches/updates/features, and just update the input when they're ready to?

Exactly, that's the typical situation in general with `nixpkgs`. Ideally, I think in the context of Ethereum.nix whenever a stable release is released, we should update the package definition quickly (hopefully automatically). If we provide tooling like [nixpkgs-versions](https://github.com/lazamar/nix-package-versions) we can help in the discoverability of when certain packages hit certain versions, in case the user might decide to pin specific pkgs for whatever reason. I'm happy to discuss more of this in another issue to nail it down.

> > Bonus points if we can cache the builds somehow (cachix?).
> 
> @shazow I _think_ this is already happening using the `nix-community` cache (the substituter that gets added when you build the `flake`) though I could be wrong, I haven't looked too closely! I originally opened PRs at nixpkgs so we could get this and `reth` in the default nixos cache, but neither have had any action yet 😅 We can probably move a bit quicker here anyways?

The main idea of this repository is to have fast access to the latest versions of packages. Period. It's a small community effort, there's no necessity for big coordination efforts like `nixpkgs`.

Also, for @shazow let's proceed then to join efforts in this repo!

aldoborrero (approved):

PR #464, opened 2024-02-12
chore: add foundry, lighthouse, wake and reth to blacklist for updating

selfuryon:
add foundry, lighthouse, wake and reth to blacklist for updating. After this change update script is working: https://github.com/nix-community/ethereum.nix/pull/466

P.S.: I'm thinking about rewriting the update script to work like in nixpkgs, where the update script runs `passthru.updateScript` instead of a hardcoded updater. I tried to use nixpkgs `update.nix` (got the idea from [here](https://discourse.nixos.org/t/how-can-i-run-the-updatescript-of-personal-packages/25274/3)) but I'm still struggling to make it work

### DIFF
--- a/.github/workflows/update-packages.yaml
+++ b/.github/workflows/update-packages.yaml
@@ -17,7 +17,7 @@ jobs:
         with:
           # TODO: remove nethermind after we fix build for them
           # TODO: remove mev-boost after they make a new release tag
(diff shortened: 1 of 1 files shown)


aldoborrero:
@selfuryon I started doing that this morning on the config-update branch (have a look at it)

aldoborrero (approved):

PR #473, opened 2024-02-16
add vouch-unstable

zimbatm:
This is needed mostly during hard forks where unstable software is
needed to test things on test networks like Holesky.


zimbatm:
This observation is also true for the other clients. Is that a pattern that we want across the board?

aldoborrero:
We can certainly add this pattern (I'm working on PR #468 which will ease auto-updating these releases quite efortless)

aldoborrero (approved):

PR #476, opened 2024-02-19
vouch: remove duplication

zimbatm:
If we want to maintain multiple versions of packages, how do we reduce
the duplication of definitions?

### DIFF
--- a/pkgs/vouch/builder.nix
+++ b/pkgs/vouch/builder.nix
@@ -1,0 +1,43 @@
+{
+  buildGoModule,
+  mcl,
+  bls,
+}: let
+  builder = {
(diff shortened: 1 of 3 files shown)


zimbatm:
Exploring this as a pattern, but it can take many different forms. The builder could be attached to a per-system `lib`. One thing I wanted to do is have the builder be easily extensible so it can override any of the original build attributes.

zimbatm:
I was playing with making the builder more easily overridable, but then it's more lines of code

aldoborrero:
I'll proceed to close this stale PR (feel free to open it again @zimbatm if you feel it should be open)

PR #489, opened 2024-03-01
mev-boost: 1.6.0 -> 1.7

jhvst:
Dencun update, tested that binary starts.

On the [release page](https://github.com/flashbots/mev-boost/releases/tag/v1.7), the authors recommend Go `1.22`. However, that is not package on Nix yet. Despite the recommendation, the program seems to compile and start fine.

### DIFF
--- a/pkgs/mev-boost/default.nix
+++ b/pkgs/mev-boost/default.nix
@@ -5,16 +5,16 @@
 }:
 buildGoModule rec {
   pname = "mev-boost";
-  version = "1.6.0";
+  version = "1.7.1";
 
   src = fetchFromGitHub {
     owner = "flashbots";
     repo = "${pname}";
     rev = "v${version}";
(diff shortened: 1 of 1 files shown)


ekimber:
The flashbots release is 1.7 rather than 1.7.1

jhvst:
Right, the page I linked is for 1.7 which is an alias of 1.7.0, but the release was followed a day later with 1.7.1. I thought it would make sense to just take the most recent version.

jhvst:
On a further thought, if there is a chance that someone missed a MEV block because of a minor version bump (I certainly don't want that), it seems more reasonable to have this at `1.7` just in case, or at least until nixpkgs is upgraded in regards to Go.

ekimber:
Ah, I hadn't noticed the 1.7.1 tag. It seems to me a little unorthodox to upgrade a go major version in a patch version increment.

selfuryon (approved):
lgtm

PR #498, opened 2024-03-09
rotki-bin: init at 1.32.1

mitchmindtree:
I had a quick go at building it from source following [the guide][1], then decided on just wrapping the appimage after running into some python build issues related to the `ruff` linter that was being invoked in a build phase for some reason. I'm sure it's doable with some patching if someone's keen enough, but this at least makes the package available for now.

[1]: https://rotki.readthedocs.io/en/latest/installation_guide.html#build-from-source

### DIFF
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -58,6 +58,7 @@
       prysm = callPackage ./prysm {inherit bls blst;};
       reth = callPackageUnstable ./reth {};
       rocketpool = callPackage ./rocketpool {};
+      rotki-bin = callPackage ./rotki-bin {};
       sedge = callPackage ./sedge {inherit bls mcl;};
       slither = callPackage ./slither {};
       ssvnode = callPackage ./ssvnode {inherit bls mcl;};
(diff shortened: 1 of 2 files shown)


aldoborrero:
I'll build it from sources, but this first version is good enough :)

aldoborrero (approved):
LGTM

jhvst:
The open PR on nixpkgs has some extra packages added to path: https://github.com/NixOS/nixpkgs/pull/281781/files#diff-43d2233c3f4a035c0efd6110ee4189d28f2260ae102c8a3a2ca12ace792b73a4R44-R47

Any idea if these are necessary?

mitchmindtree:
Ahh I didn't see that PR @jhvst - I just checked that the application GUI ran and it was looking good. I'm unsure whether or not there are any deeper runtime issues that might be caused without those on linux?

Either way, that nixpkgs PR appears to also support darwin, so would be great to see that extra support landed either in nixpkgs or here.

PR #503, opened 2024-03-19
eth-validator-watcher: init at 0.7.1

ekimber:
Kilnfi's validator watcher.  May be interesting to anyone who wants to roll their own notifier, for upcoming block proposals etc. rather than relying on an external service such as beaconcha.in
I have tried this with lighthouse and it works (using the generic beacon API)


aldoborrero:
@ekimber nice addition!

aldoborrero (approved):

PR #507, opened 2024-04-01
snarkjs: init at 0.7.3

jhvst:
This is related to Ethereum as it is a dependency of https://eth-store.w3eth.io 

I will package EthStorage later, using this as a dependency.


aldoborrero (approved):
LGTM

PR #511, opened 2024-04-14
besu: use jemalloc if possible

rolfschr:
This PR recycles my (maybe-merged-one-day) [PR from nixpkgs](https://github.com/NixOS/nixpkgs/pull/282523). A little bit of information can be found there.

However, this PR introduces tests in the package definition itself. I expect change requests here since I haven't seen any other packages that defines `.passthru.tests` so I don't think this is a fit currently. From a cursory look into the code/issues/PRs, I guess you would want to have dedicated module tests. I'm happy to adapt the code in any direction (keeping tests, though), just let me know what you expect.

### DIFF
--- a/pkgs/besu/default.nix
+++ b/pkgs/besu/default.nix
@@ -1,11 +1,14 @@
 {
   fetchurl,
+  jemalloc,
   jre,
   lib,
   makeWrapper,
+  runCommand,
   stdenv,
+  testers,
 }:
-stdenv.mkDerivation rec {
+stdenv.mkDerivation (finalAttrs: rec {
   pname = "besu";
   version = "24.1.2";
 
@@ -14,6 +17,7 @@ stdenv.mkDerivation rec {
     hash = "sha256-CC24z0+2dSeqDddX5dJUs7SX9QJ8Iyh/nAp0pqdDvwg=";
   };
 
+  buildInputs = lib.optionals stdenv.isLinux [jemalloc];
   nativeBuildInputs = [makeWrapper];
 
   installPhase = ''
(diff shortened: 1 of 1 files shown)


aldoborrero:
@rolfschr first of all, thanks for your contribution.

Secondly, I've been slowly getting back to action as I've been recently a father, so apologies for not getting back to you earlier.

Related to this:

> However, this PR introduces tests in the package definition itself. I expect change requests here since I haven't seen any other packages that defines .passthru.tests so I don't think this is a fit currently. From a cursory look into the code/issues/PRs, I guess you would want to have dedicated module tests. I'm happy to adapt the code in any direction (keeping tests, though), just let me know what you expect.

It's OK to have them defined there and it's not uncommon on nixpkgs. Right now, we're only using the `flake.checks` as a way of ensuring that everything works as expected:

```nix
checks =
          # merge in the package derivations to force a build of all packages during a `nix flake check`
          // (with lib; mapAttrs' (n: nameValuePair "package-${n}") (filterAttrs (n: _: ! builtins.elem n ["docs"]) self'.packages))
          # mix in tests
          // config.testing.checks;
```

Maybe, we can iterate the list of packages and if it has defined the `passtrhu.tests` attrset, we can include them on the list of things to check.

What do you think?

Eventually, we will refine our infra to be more similar to `nixpkgs`.

rolfschr:
> Secondly, I've been slowly getting back to action as I've been recently a father, so apologies for not getting back to you earlier.

Congrats!

I think it would be good to keep the `passthru.tests` and adapt testing infra later in a separate PR which will always run existing `passthru.tests`.

As for this PR, I realized also that it covers the possibility of `system` being darwin which you do not seem to target. I can keep it as is or remove specific stuff.

aldoborrero:
@rolfschr thanks!

> As for this PR, I realized also that it covers the possibility of system being darwin which you do not seem to target. I can keep it as is or remove specific stuff.

Just add `darwin` to the bottom of the `meta.platforms` and it will be supported :)

rolfschr:
My point is: Please tell me what you (people) want/need and I adapt so this can be merged.

pkgs/besu/default.nix
```diff
@@ -32,4 +57,4 @@ stdenv.mkDerivation rec {
     platforms = ["x86_64-linux"];
```
aldoborrero, line 57: Add `"aarch64-darwin" here.

Ensure it builds and you're able to run it with `nix run .#packages.x86_64-linux.besu` (same should be for `darwin`) and that also displays the `jemalloc` string.

Also, wait for CI to have checks green for `besu`.

And done.

aldoborrero (commented):
See the comment I left.

rolfschr:
Okay, I checked `nix build --print-build-logs .#besu.tests.version`, `nix build --print-build-logs .#besu.tests.jemalloc` & `nix run .#packages.x86_64-linux.besu`. All runs as expected. I do not have `darwin` at hand. This looks done & fine from my point of view. Feel free to merge this.

aldoborrero (approved):

PR #517, opened 2024-05-16
fix eval

zimbatm:
Trying to fix the eval issue in the default branch


zimbatm:
@aldoborrero not sure the root cause but do you want to keep the wake program?

zimbatm:
The issue is probably related to a new version of nix-eval-jobs, or some of its parameters have changed in buildbox-nix

aldoborrero:
Bye for now Wake!

PR #520, opened 2024-05-26
HELP NEEDED: hardhat: init at WIP

marijanp:
I'm not familiar with pnpm or node, I could use some help on how to properly implement the installPhase.


aldoborrero:
@marijanp Packaging `hardhat` is a big effort that I highly suspect will require significant maintenance to keep it up to date with the latest development (or at least that has been my experience packaging node applications). Given that this PR hasn't received significant changes in the past few months, I'll consider it stale, so I'll proceed with closing it.

Let me know if it should be re-opened otherwise.

PR #521, opened 2024-05-28
nimbus: 24.3.0 -> 24.5.1

catwith1hat:
`nix build '.#nimbus'` works for me. Also the validator binary does its job, which is the only binary I am using from this package.


selfuryon (approved):
LGTM

PR #531, opened 2024-07-13
Implement Reth service module

scottbot95:
I tested this against Reth 1.0 (from #530), however I don't think the command line has changed since the current imported version (0.2.0-beta.6) so I opted to implement it against `main` instead of on top of my other PR.

I chose not to implement and integration test with a nixos vm because I felt it is mostly likely unecessary due to the extensive test coverage within Reth itself.

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -18,6 +18,7 @@
       ./prysm-beacon
       ./prysm-validator
       ./restore
+      ./reth
     ];
   };
 }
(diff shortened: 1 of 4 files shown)


aldoborrero:
@scottbot95 and we will migrate soonish to Reth 1.0, so even if it doesn't work at all, it's not a problem to merge it now.

aldoborrero (changes requested):
@scottbot95 only minor thing:

- Remove mdDoc as those are effectively deprecated.

scottbot95:
Oh I didn't realize `mdDoc` isn't needed anymore. It got included since I just copied from another module, but I can remove them.

aldoborrero (approved):

PR #536, opened 2024-07-21
rocketpoold: init at 1.13.7

ekimber:
Need the daemon to write a rocketpool module that works without docker.
Perhaps it could be included in the rocketpool (client) package? But the build environment conflicts and I'm not sure how best to overcome that, other than just package it separately.


aldoborrero (approved):
LGTM

PR #540, opened 2024-08-06
Fix nethermind module

ekimber:
Module was missing a backslash and included a systemd option that causes a core dump, presumably because it isn't compatible with JIT elements in the dotnet runtime


ekimber:
It looks like this is needed for the nimbus module too.

aldoborrero (approved):

aldoborrero:
Thanks @ekimber

PR #548, opened 2024-09-25
Separate nethermind JSON RPC Host and EngineHost options

ekimber:
Separate the RPC host and engine host options so that they can, for example, bind to different interfaces

### DIFF
--- a/modules/nethermind/default.nix
+++ b/modules/nethermind/default.nix
@@ -126,7 +126,7 @@ in {
               description = "Nethermind Node (${nethermindName})";
 
               environment = {
-                WEB3_HTTP_HOST = cfg.args.modules.JsonRpc.EngineHost;
+                WEB3_HTTP_HOST = cfg.args.modules.JsonRpc.Host;
                 WEB3_HTTP_PORT = builtins.toString cfg.args.modules.JsonRpc.Port;
               };
 
--- a/modules/nethermind/args.nix
+++ b/modules/nethermind/args.nix
@@ -68,6 +68,12 @@ with lib; {
         description = "Defines whether the JSON RPC service is enabled on node startup.";
       };
 
+      Host = mkOption {
+        type = types.str;
+        default = "127.0.0.1";
+        description = "Host for JSON RPC calls.";
+      };
+
       Port = mkOption {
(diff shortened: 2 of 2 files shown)


aldoborrero:
`EngineHost` then is not being used at all, if I'm not mistaken, isn't it?

ekimber:
> @ekimber `EngineHost` is not being used at all. If I'm not mistaken, isn't it?

It is used in the ExecStart command, like this for example:
`ExecStart=/nix/store/dl6i36513qymiggrksyn75nvcxmgx6zp-nethermind-1.25.4/bin/nethermind --datadir %S/nethermind-mainnet \
--JsonRpc.JwtSecretFile %d/jwtsecret \
--log INFO \
--HealthChecks.Enabled true \
--JsonRpc.Enabled true \
--JsonRpc.EngineHost 127.0.0.1 \
--JsonRpc.EnginePort 8551 \
--JsonRpc.Host 0.0.0.0 \
--JsonRpc.Port 8545 \
`

So you can expose the RPC (with all the security caveats) without exposing the engine

modules/nethermind/default.nix
```diff
@@ -126,7 +126,7 @@ in {
               description = "Nethermind Node (${nethermindName})";
 
               environment = {
-                WEB3_HTTP_HOST = cfg.args.modules.JsonRpc.EngineHost;
+                WEB3_HTTP_HOST = cfg.args.modules.JsonRpc.Host;
```
bbjubjub2494, line 129: By the way what exactly are those environment variables consumed by?

ekimber: The backup module needs them. Though I have never used that module myself

bbjubjub2494 (approved):
I think the change is good. Please rebase, it will probably build

ekimber:
Yes it is still relevant! If it gets merged I can finally delete my branch and get the nethermind box back onto main 🙏

aldoborrero:
@ekimber @bbjubjub2494 thanks! Merged! Ensure I'm one of the PR reviewers so I get notified on Github. I'll add a `codeowners` file in a new PR to have it automatically done.

PR #552, opened 2024-10-21
nimbus: 24.6.0 -> 24.9.0

ekimber:
Update to latest nimbus. 

I see they have a flake.nix in the nimbus repo now so could try using this as an input.

### DIFF
--- a/pkgs/nimbus/default.nix
+++ b/pkgs/nimbus/default.nix
@@ -2,62 +2,53 @@
   fetchFromGitHub,
   lib,
   stdenv,
-  lsb-release,
   which,
   cmake,
   pkgs,
   writeShellScriptBin,
   buildFlags ? ["nimbus_beacon_node" "nimbus_validator_client"],
-}: let
-  nim1 = pkgs.nim-unwrapped-1.overrideAttrs rec {
-    version = "1.6.18";
-    src = pkgs.fetchurl {
-      url = "https://nim-lang.org/download/nim-${version}.tar.xz";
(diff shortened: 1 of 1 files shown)


pkgs/nimbus/default.nix
```diff
@@ -2,62 +2,53 @@
...
+  fakeGit = writeShellScriptBin "git" "echo ${version}";
+
+  # Dunno why we need `lsb-release`, looks like for nim itself
+  # without `which` it can't find gcc
+  nativeBuildInputs = [fakeGit pkgs.nim which cmake];
+  enableParallelBuilding = true;
+
+  NIMFLAGS = "-d:disableMarchNative -d:release";
+
+  makeFlags = ["USE_SYSTEM_NIM=1"];
```
jakubgs, line 32: Keep in mind that using Nim provided by Nix is not supported, since `nimbus-eth2` is tightly coupled with compiler version.

jakubgs:
Also yes, using our own flake would probably be a good idea, we'd be happy to help with integrating things if issues arise.

pkgs/nimbus/default.nix
ekimber: OK, I looked for the compiler version in the source repo because it had clearly moved to nim 2.x but I couldn't find it. Though I now see it some commit messages referencing 2.0.10. I will fix this in our package.

ekimber:
> Also yes, using our own flake would probably be a good idea, we'd be happy to help with integrating things if issues arise.

Thanks, I will give it a go

ekimber:
Superseded by #557

PR #545, opened 2024-09-20
ci(Mergify): configuration update

zowoq:
This change has been made by @zowoq from the Mergify config editor.

Once this is merged you can re-add the `nix-community-buildbot` topic.

https://github.com/orgs/nix-community/discussions/1457

PR #556, opened 2024-10-30
chore(ci): small enhancement

selfuryon:
I found out that we had some problems with the current mergify configuration, updates it to the latest one. Is it ok?
It also contains changes from https://github.com/nix-community/ethereum.nix/pull/545


selfuryon:
After I will try to reenable buildbot (I just realized, it doesn't work)

brianmcgee (approved):

zowoq:
> After I will try to reenable buildbot 

I've reenabled it.

PR #557, opened 2024-11-01
nimbus: 24.6.0 -> 24.10.0

ekimber:
Build the nimbus package from the source repo flake. Updating is super easy now, just need to update the flake input :)

Not sure if it's safe to use follows on nixpkgs here or not so I've left it out but of course there is an impact on evaluations.

I am running this on a live validator. It seems to be working fine.


aldoborrero:
@ekimber I'm still unsure if it's a good idea to add support now to nimbus here if there's an official flake with proper support from the project. We only provide a convenience mechanism for end users not having to add that flake.

Opinions?

ekimber:
> @ekimber I'm still unsure if it's a good idea to add support now to nimbus here if there's an official flake with proper support from the project. We only provide a convenience mechanism for end users not having to add that flake.
> 
> Opinions?

I completely agree it's preferable to use the official project flake directly. I've already done that in my own nixos configuration. However, I do see value in the `ethereum.nix` Nimbus module and would like to continue using it. It needs a Nimbus package to 'just work' and the official package is better than an unofficial one. Either way, some documentation pointing to the official flake would be helpful.

bbjubjub2494:
There are two issues with this approach:
1. downloading all the submodules at eval time takes forever
2. There is a mismatched hash in the latest version. https://github.com/status-im/nimbus-eth2/blob/4e440277cf8a3fed72f32eb2f01fc5e910ad6768/nix/nimble.nix#L11 More generally if we have an issue with the upstream nix expression it's very ugly to patch

I don't know what the best approach is though.

ekimber:
>     2. There is a mismatched hash in the latest version. https://github.com/status-im/nimbus-eth2/blob/4e440277cf8a3fed72f32eb2f01fc5e910ad6768/nix/nimble.nix#L11 More generally if we have an issue with the upstream nix expression it's very ugly to patch

They are looking into why the CI didn't catch that mismatch. The flake could be changed so that the version is pinned here to 24.10.0, with the benefit that someone using ethereum.nix would never see that issue as we could skip straight to the upcoming version where it will be fixed.

ekimber:
Superseded by #568

Issue #558, opened 2024-11-18
mev-boost module startup test is not working


ekimber:
### Bug description

The network option needs to be changed from goerli to sepolia because goerli is no longer supported. However, when I try this myself I get an invalid relay URL error citing "duplicate entry" which is odd. 

### Steps to reproduce

1. Failing test is currently breaking the buildbot builds

### Before submitting

- [X] I have attached links to [the documentation](https://nix-community.github.io/ethereum.nix), and possibly related [issues](https://github.com/nix-community/ethereum.nix/issues) and [discussions](https://github.com/nix-community/ethereum.nix/discussions).

ekimber:
Fixed by #559

PR #559, opened 2024-12-16
fix mev-boost module option generation and network options

ekimber:
Contains the following changes:

- Remove goerli as it is retired and no longer a valid network option.
- Fix the cli option generation so that it works with the change in mev-boost v1.8 to use the urfave cli library.

Resolves issue #558 
The mev-boost test passes for me now.


zimbatm (approved):
nice. thank you

PR #560, opened 2024-12-19
reth: 0.2.0-beta.6 -> 1.1.4

### LINKED ISSUE #526: Update reth to 1.0
### Context

Reth 1.0 has been released: https://www.paradigm.xyz/2024/06/reth-prod

### Description

The current package is still on the alpha version

### Before submitting

- [X] I have verified that my idea is a change request and not a bug report

bbjubjub2494:
Latest version, ~1 week ago.

Fixes #526 

Acknowledging #530. I don't have an opinion on using crane. Currently the "official" way just works so this is orthogonal.

### DIFF
--- a/pkgs/reth/default.nix
+++ b/pkgs/reth/default.nix
@@ -1,9 +1,7 @@
 {
-  darwin,
   fetchFromGitHub,
   lib,
   rustPlatform,
-  stdenv,
 }:
 rustPlatform.buildRustPackage rec {
   pname = "reth";
@@ -24,14 +22,6 @@ rustPlatform.buildRustPackage rec {
     rustPlatform.bindgenHook
   ];
 
-  # `x86_64-darwin` seems to have issues with jemalloc
-  buildNoDefaultFeatures = true;
-  buildFeatures = lib.optional (stdenv.system != "x86_64-darwin") "jemalloc";
-
-  buildInputs = lib.optionals stdenv.isDarwin [
-    darwin.apple_sdk.frameworks.Security
-  ];
-
   # Some tests fail due to I/O that is unfriendly with nix sandbox.
   checkFlags = [
     "--skip=builder::tests::block_number_node_config_test"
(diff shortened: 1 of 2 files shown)


bbjubjub2494:
Acknowledging that the dependency bump broke some stuff

aldoborrero:
@bbjubjub2494 Let's drop support for `darwin` for now. Also, would you mind reviewing this?

https://buildbot.nix-community.org/#/builders/1662/builds/24

bbjubjub2494:
> @bbjubjub2494 Let's drop support for `darwin` for now. Also, would you mind reviewing this?
> 
> https://buildbot.nix-community.org/#/builders/1662/builds/24

I am clueless as far as Darwin intrinsics go. Was that what I was supposed to review?

aldoborrero:
On #530, we discussed dropping support to `darwin` for `reth`. We only need to remove from the `meta.platforms` these two: `"aarch64-darwin" "x86_64-darwin"`.

bbjubjub2494:
closed in favor of #568

pkgs/reth/default.nix
```diff
@@ -1,42 +1,27 @@
...
-    };
   };
 
   nativeBuildInputs = [
     rustPlatform.bindgenHook
   ];
 
-  # `x86_64-darwin` seems to have issues with jemalloc
-  buildNoDefaultFeatures = true;
-  buildFeatures = lib.optional (stdenv.system != "x86_64-darwin") "jemalloc";
```
jhvst, line 34: Line 34 seems to have a regression, no? In the commit to remove darwin support, I think this line was accidentally removed, which was used to enable `jemalloc` on _any other_ platform than x86 darwin.

bbjubjub2494: Good catch! I naively assumed that the no default features flag was here because of the next line, but I did not check what the default features were. If there are any besides `jemalloc` though we actually enabled them...

bbjubjub2494: the only default feature is in fact jemalloc. https://github.com/paradigmxyz/reth/blob/1bf8d504616752dc4cc70071bc240c38b7c843e5/bin/reth/Cargo.toml#L99

Anyway we can keep the vanilla approach.

PR #561, opened 2024-12-19
foundry: nightly-0688b5a -> nightly-8a08a3a

bbjubjub2494:
~~waiting on https://github.com/NixOS/nixpkgs/pull/361878 for rust 1.83~~

PR #562, opened 2025-01-04
nethermind: 2.25.4 -> 2.30.3

bbjubjub2494:
On top of the version bump, I had to fight the derivation a bit.

With `projectFile = "src/Nethermind/Nethermind.sln";`, there are some errors. The [Nixpkgs Manual, Packaging a dotNET application](https://nixos.org/manual/nixpkgs/stable/#packaging-a-dotnet-application) advises against putting solution files in this variable.

~~The .nix format for nuget deps is deprecated so I took the opportunity to upgrade.~~ nevermind (ha!)

One slightly annoying thing is that one of the dependencies, `System.CommandLine v2.0.0-beta4.24517.1` is hosted in the `https://pkgs.dev.azure.com` repository which doesn't work automatically with `nuget-to-nix`.


bbjubjub2494:
TODO(me) fix checks

ekimber:
Nice work getting this to compile. I could not get fetch-deps working. I would like to move to 24.11 for this because the current dotnet sdk 9.0.0 is a preview release.

aldoborrero:
Nice work @bbjubjub2494 would you mind running `treefmt`?

bbjubjub2494:
Maybe a PR centered around bumping flake inputs is in order? Then we can iterate with content from this as well as #561

aldoborrero:
Ideally, we should bump automatically all inputs with a GitHub Action or something that allows us to keep up without manual intervention.

If [I'm not mistaken renovate does support updating flake inputs](https://github.com/renovatebot/renovate/blob/50e53440f4e16726cc1836280c2893496ed4f48f/lib/modules/manager/nix/artifacts.ts#L38).

PR #563, opened 2025-01-04
geth: expose blsync

bbjubjub2494:
We build blsync so might as well make it official


aldoborrero:
@bbjubjub2494 would you mind signing the commit? Otherwise LGTM

aldoborrero (approved):

PR #568, opened 2025-01-08
Update flake dependencies

bbjubjub2494:
the plan is to do `nix flake update --commit-lock-file` and then get green on buildbot. Incorporates #560 #561 and #562. As discussed in #560 we drop support for reth on Darwin. I haven't touched the `nixos-24.05` input but this should be done asap

### DIFF
--- a/pkgs/nimbus/default.nix
+++ b/pkgs/nimbus/default.nix
@@ -1,63 +1,19 @@
 {
+  applyPatches,
   fetchFromGitHub,
-  lib,
-  stdenv,
-  lsb-release,
-  which,
-  cmake,
   pkgs,
-  writeShellScriptBin,
-  buildFlags ? ["nimbus_beacon_node" "nimbus_validator_client"],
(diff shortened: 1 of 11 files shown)


pkgs/nimbus/default.nix
```diff
@@ -1,63 +1,19 @@
...
-      homepage = "https://github.com/status-im/nimbus-eth2";
-      license = lib.licenses.asl20;
-      mainProgram = "nimbus_beacon_node";
-      platforms = ["x86_64-linux"];
-    };
-  }
+    patches = [./fix-hash.patch];
+  };
+in
+  import "${src}/nix" {inherit pkgs targets;}
```
bbjubjub2494, line 19: ok somehow I landed on this approach which I guess avoids some of the weirdness of directly importing flakes around, while still reusing upstream nix code. It's technically IFD, but from a FOD, and doesn't trigger if you don't use Nimbus.

PR #572, opened 2025-02-03
fix lighthouse args.beacon-nodes

bbjubjub2494:
the condition is backwards. As a result:
- If you set the option to `[]`, it fails at runtime with `error: a value is required for '--beacon-nodes <NETWORK_ADDRESSES>' but none was supplied`.
- if you set is to `[my-beacon-enpoint]`, it uses the default value (or lighthouse-beacon value) instead.


aldoborrero (approved):
LGTM. Thanks @bbjubjub2494

PR #573, opened 2025-02-13
foundry: nightly -> 1.0.0

bbjubjub2494:
https://www.paradigm.xyz/2025/02/announcing-foundry-v1-0

It seems sensible that the default derivation should be stable at this point. There's also [foundry.nix](https://github.com/shazow/foundry.nix) for nightlies


selfuryon (approved):
LGTM

PR #574, opened 2025-02-19
update(pkgs): Update geth, nethermind, nimbus, web3signer and mev-boost packages

MartinNikov:
* chore(flake.lock): Update all Flake inputs (2025-03-04)
* nethermind: 1.30.3 > 1.31.3
* geth: 1.14.13 > 1.15.4
* mev-boost: 1.8 > 1.9-rc2
* web3signer: 24.2.0 > 25.2.0
* nimbus: 24.12.0 > 25.3.0


pkgs/nimbus/default.nix
```diff
@@ -5,16 +5,13 @@
...
-      hash = "sha256-DBvsnGr91a69eCj1hAeoVOpxas5rfaT36rIxWEmvIVg=";
-      fetchSubmodules = true;
-    };
-    patches = [./fix-hash.patch];
+  version = "25.2.0";
+  src = fetchFromGitHub {
+    owner = "status-im";
+    repo = "nimbus-eth2";
+    rev = "v${version}";
+    hash = "sha256-aPyfbQ1983jaWL2XI0yyLHXYegIPT7ugPbFfid5pkfo=";
```
bbjubjub2494, line 13: It looks like that hash doesn't match on my machine. I hope it's not nondeterministic :slightly_frowning_face:

MartinNikov: Although it builds for me, I tested it on a different computer and got a different hash. After using that, there ought to be no issues.

bbjubjub2494 (changes requested):
It looks like the dependency bump changed the dep hash for blutgang. https://buildbot.nix-community.org/#/builders/1605/builds/60

I looked into it, it looks like the contents of the tarballs are identical, only some files appear in a different order in the tarball.

Please also run treefmt at the end

pkgs/nethermind/nuget-deps.nix
```diff
@@ -1,30 +1,28 @@
-# This file was automatically generated by passthru.fetch-deps.
-# Please dont edit it manually, your changes might get overwritten!
```
bbjubjub2494, line 2: This file is just a little bit strange. This comment is missing, and the contents don't exactly match what I can generate automatically. There are some extra entries. Why is that?

bbjubjub2494 (changes requested):

pkgs/nethermind/nuget-deps.nix
MartinNikov: Here, `System.CommandLine v2.0.0-beta4.24528.1`, which was obtained from the Azure source, was the cause of the nuget-to-nix failure. For this reason, I use nuget-to-nix manually and obtain all of Nethermind's dependencies except for this one. I then manually added it to the file.

pkgs/nimbus/default.nix
PetarKirov: I've [investigated](https://github.com/metacraft-labs/ethereum.nix/pull/43#issuecomment-2673941274) myself the difference and found out the following:

@MartinNikov used `nurl` (which in turn uses `nix flake prefetch`) and got the first hash, while the other hash is obtained when you put fake hash in `fetchFromGitHub`.

The difference between the two is from Windows-specific files that were committed with CRLF in one of the submodules. `nix flake prefetch` normalizes them to LF, while `fetchFromGitHub` does not.

bbjubjub2494 (approved):
Fine with me, still needs a treefmt

yakimant:
Pectra release is today on holesky:
https://blog.ethereum.org/2025/02/14/pectra-testnet-announcement#client-releases
Geth/Nethermind upgrades in this PR are mandatory.

yakimant:
Pectra for Holesky didnt work well, so more upgrades needed for both geth and nethermind.

yakimant:
- https://github.com/ethereum/go-ethereum/pull/31247
- https://github.com/NethermindEth/nethermind/releases/tag/1.31.1

bbjubjub2494:
The build should be fixed with facc2b32d7cb34a2865e1bc3d075026c35f23733

MartinNikov:
Finally, all checks have passed. Regarding `mdformat`, I used the old version because the newest one was broken. Is that fine with you, @bbjubjub2494?

bbjubjub2494:
> Finally, all checks have passed. Regarding `mdformat`, I used the old version because the newest one was broken. Is that fine with you, @bbjubjub2494?

I'm not an authority on mdformat, but afaik there's no other known solution so no objections

bbjubjub2494 (dismissed):
LGTM

ekimber:
Not sure about using release candidates

MartinNikov:
Why not? The hard fork required those versions.

aldoborrero:
Hi everybody,

just a couple of things:

1. We can disable `mdformat` if it causes more issues. I was initially maintaining the package on `nixpkgs` but haven't had more time to do it properly, and there are some community members who stepped up. So, if it brings any issues, we can just disable the flag until it's fixed.
2. In terms of using `rc` I'm not against it for Consensus or Execution clients as long as it makes sense for new network forks. I guess there's now a necessity of having multiple versions in place (i.e, the stable and the rc).
3. @MartinNikov can you [review this failure](https://buildbot.nix-community.org/#/builders/1646/builds/73)?

bbjubjub2494:
As agreed with @MartinNikov I am taking over this PR and splitting it

- [x] nethermind: 1.30.3 > 1.31.3 #599
- [ ] geth: 1.14.13 > 1.15.4
- [ ] mev-boost: 1.8 > 1.9-rc2
- [ ] web3signer: 24.2.0 > 25.2.0
- [x] nimbus: 24.12.0 > 25.3.0 #598

bbjubjub2494:
all changes have been merged in equivalent PRs

PR #592, opened 2025-03-24
fix: apps discrepancies in docs

bbjubjub2494:
Some apps listed in the table do not match what's in the flake

Depends on #590


bbjubjub2494 (commented):
Can't approve my own PR

PR #608, opened 2025-05-06
Updated prysm and geth to versions that support pectra mainnet

JulianGodd:
prysm 5.3.0 -> 6.0.1
geth 1.15.9 -> 1.15.11

Updates for the Pectra mainnet changes.

Sorry if this isn't the correct way of moving this forward.

The build is failing due to the following error with prysm:
```
Building subPackage ./cmd/beacon-chain
# github.com/ethereum/c-kzg-4844/v2/bindings/go
vendor/github.com/ethereum/c-kzg-4844/v2/bindings/go/main.go:5:11: fatal error: ckzg.c: No such file or directory
    5 | // #include "ckzg.c"
```

I'm able to build from the prysm repo separately using `go build -v ./...` although I lack the familiarity with go to differentiate this from how the nix build works.

### DIFF
--- a/pkgs/by-name/ge/geth/default.nix
+++ b/pkgs/by-name/ge/geth/default.nix
@@ -21,17 +21,17 @@
 in
   buildGoModule rec {
     pname = "geth";
-    version = "1.15.9";
+    version = "1.15.11";
 
     src = fetchFromGitHub {
       owner = "ethereum";
       repo = "go-ethereum";
       rev = "v${version}";
-      hash = "sha256-zD4uJJPVHiJ44+pBALQ4tEvYda3S7SIhn0ySyT6BSAQ=";
+      hash = "sha256-2XGKkimwe9h8RxO3SzUta5Bh2Ooldl2LiHqUpn8FK7I=";
     };
 
     proxyVendor = true;
-    vendorHash = "sha256-1FuVdx84jvMBo8VO6q+WaFpK3hWn88J7p8vhIDsQHPM=";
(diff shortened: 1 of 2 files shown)


JulianGodd:
I don't think we need to switch to bazel here as `go build -v ./...` works (unless this somehow uses bazel under the hood).

The file ckzg.c (v2.1.1) appears to be just a rename of c_kzg_4844.c (v1.0.0) and hasn't changed directory.

While these c files exist in the nix store, for both cases they exist in different directories than the ones listed in the error and the derivations that have them don't contain go bindings. Given that in the go bindings for 2.1.1 and 1.0.0 the path is relative I'm surprised these aren't either both working or both broken.

Do you know how I can view the exact commands nix uses to build here so I can replicate the error?

bbjubjub2494:
> I don't think we need to switch to bazel here as `go build -v ./...` works (unless this somehow uses bazel under the hood).

It doesn't seem to me like `go build` does that. Otherwise good, let's not touch Bazel at all if possible

> The file ckzg.c (v2.1.1) appears to be just a rename of c_kzg_4844.c (v1.0.0) and hasn't changed directory.

Nice

> While these c files exist in the nix store, for both cases they exist in different directories than the ones listed in the error and the derivations that have them don't contain go bindings. Given that in the go bindings for 2.1.1 and 1.0.0 the path is relative I'm surprised these aren't either both working or both broken.

:woman_shrugging: 

> Do you know how I can view the exact commands nix uses to build here so I can replicate the error?

If I understood what you want, `nix-shell -A packages.x86_64-linux.prysm` should drop you in a shell environment that reflects the build sandbox (not perfectly!) From there, you can run the build phases manually

PR #609, opened 2025-05-07
erigon: 2.61.3 -> 3.0.3

bohendo:
Not sure why we didn't already have a 24.11 nixpkgs input, we have old stable ones and unstable. I added the latest stable version and, in the future, I think a lot of packages could migrate to using the same nixpkgs 24.11 stuff that erigon needed & maybe we can remove some of the older nixpkgs. But that's not the focus of this PR.

The [erigon blog](https://erigon.tech/releasing-erigon-v3-0-0/) says this upgrade is required for Pectra. But beware: requires a complete resync if upgrading from the 2.x.x series

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -131,6 +131,22 @@
         "type": "github"
       }
     },
+    "nixpkgs-2411": {
+      "locked": {
+        "lastModified": 1731603435,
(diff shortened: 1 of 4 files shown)


bbjubjub2494 (approved):
LGTM per se. There needs to be some maintenance done with regards to nixpkgs inputs, but I'll focus on that in #611

PR #578, opened 2025-02-22
lighthouse: 5.3.0 -> 7.0.1

bbjubjub2494:
This has support for Pectra as well as a backwards-incompatible new database schema.

Currently the build is failing because a submodule contains a bad symlink. I will patch this as soon as I have some time https://github.com/jmacd/xdelta/blob/release3_0_apl/xdelta3/py-compile

PR #610, opened 2025-05-07
lighthouse: 5.3.0 -> 7.0.1

bohendo:
I made this change before checking existing PRs, oops. We may want to prioritize PR #578 because that one deactivates a bunch of failing tests and this one deactivates all tests. [Identifying which tests needed to be skipped](https://github.com/nix-community/ethereum.nix/pull/578#issuecomment-2852005353) likely involved some long, slow tweak/checks, so hat tip to @evilbulgarian. But, by skipping all tests, this one will build faster :smiling_imp:


bbjubjub2494:
Thank you for your work! It looks like the CI caught some unused nix parameters. Can you remove them?

https://buildbot.nix-community.org/#/builders/1637/builds/136

bbjubjub2494:
@bohendo since I managed to get #578 to pass build, I will merge it instead. Besides the tests, it also manages to remove the bundled Cargo.lock and the patch. Thank you for your contribution!

PR #616, opened 2025-05-27
inputs: 24.11, unstable -> 25.05

bbjubjub2494:
Since 25.05 branched off unstable, we can probably reduce the number of inputs, at least for some time. Will it build? :crossed_fingers:


selfuryon:
I think we can close it due to merge of https://github.com/nix-community/ethereum.nix/pull/650

PR #626, opened 2025-07-20
Add user option in geth nixos module.

thalin:
I wanted to manage the user that runs the geth service directly instead of relying on DynamicUser in the systemd config. The existing geth config didn't really allow this, but adding mkDefault to User let me override it in my own config. While I was there, I thought maybe it would be useful to be able to change StateDirectory and ExecStart too, so I went ahead and wrapped those with mkDefault too - but I'm not really sure ExecStart needs it. Looking for a second opinion there.

FWIW, I wrote a little `ethOverride` function to change the user and add (most of) the settings that DynamicUser covers, which looks like this:

```
ethOverride = user: {
  DynamicUser = mkForce false;
  RemoveIPC = true;
  PrivateTmp = true;
  NoNewPrivileges = "strict";
  RestrictSUIDSGID = true;
  ProtectSystem = true;
  User = mkForce user;
};
```

Then I call the function like `systemd.services.<service-name>.serviceConfig = ethOverride myuser`.

### DIFF
--- a/modules/geth/default.nix
+++ b/modules/geth/default.nix
@@ -19,6 +19,7 @@
     flatten
     mapAttrs'
     mapAttrsToList
+    mkDefault
     mkIf
     mkMerge
     nameValuePair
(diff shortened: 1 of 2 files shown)


thalin:
This could probably be more well-supported in the module directly of course - adding a user option would be nice. However, this is most expedient for what I needed so here it is.

thalin:
OK I think I am happy with this now and think that it doesn't disrupt existing behavior. Any feedback would be appreciated!

thalin:
Closed in favor of #1239

PR #627, opened 2025-07-26
web3signer: Init module.

catwith1hat:
Here is a module that I used awhile ago, which I never commited upstream. I since moved away from web3signer. I am not 100% sure that the module is functional, but I can't see a reason why not.

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -19,6 +19,7 @@
       ./prysm-validator
       ./restore
       ./reth
+      ./web3signer
     ];
   };
 }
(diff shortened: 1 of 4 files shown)


bbjubjub2494 (changes requested):
Hi

> I am not 100% sure that the module is functional

For me personally, this would be a good reason not to merge your code, because then we get the maintenance burden for the module with no idea if it actually improves anything downstream. I will let others opine though.

Regardless thank you for making the effort to open source it, hope you understand!

bbjubjub2494:
no one expressed interest so I am going to close this for now

PR #628, opened 2025-07-26
Module for Nimbus and its validator

catwith1hat:
Two basic module definitions for Nimbus and its validator. I moved on from using this code, but it might be useful for somebody else.


bbjubjub2494:
on it #647

PR #629, opened 2025-08-10
ethstaker-deposit-cli: Init at 1.2.2

catwith1hat:
Has support for Hoodi testnet and a few other features that the official tool is lacking.


catwith1hat:
Signed the commit.

PR #643, opened 2025-08-24
chore: set up daily flake update with Hercules CI

bbjubjub2494:
There is currently no automation to automatically try to update flake inputs. hercules-ci can do it. I don't think it is able to sign its commits, which could be an issue

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -99,6 +99,29 @@
         "type": "github"
       }
     },
+    "hercules-ci-effects": {
+      "inputs": {
+        "flake-parts": [
+          "flake-parts"
+        ],
+        "nixpkgs": [
+          "nixpkgs"
+        ]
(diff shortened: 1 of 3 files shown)


aldoborrero:
@bbjubjub2494 have a look at `numtide/nix-ai-tools` as a reference for an alternative approach without using `hercules-ci`:

- https://github.com/numtide/nix-ai-tools/blob/main/.github/workflows/update-flake.yml
- https://github.com/numtide/nix-ai-tools/blob/main/.github/actions/update.sh

aldoborrero:
Ok let's give it a try. The only unclear part for me on this is the commit signature. Is it something we can configure?

bbjubjub2494:
Honestly I don't know and I can't figure it out

PR #652, opened 2025-08-28
feat: flatten (again) all packages

aldoborrero:
# Summary of Changes

This, again, introduces flattening the packages on the project that I introduced a couple of months back. The idea is to simplify inputs, so we will take advantage of this simplification soon.


selfuryon:
> @selfuryon Ah, I said the opposite. Yep, don't worry.

Looks like we made comments at the same time) Actually I liked to incorporate it here more, let me add commit here)

selfuryon:
I added my commits here :) @aldoborrero

selfuryon (approved):
lgtm

aldoborrero:
@selfuryon it seems so!

PR #655, opened 2025-09-02
chore(ethstaker-deposit-cli): change repo url

selfuryon:
There are a new url here which we should I believe: https://github.com/ethstaker/ethstaker-deposit-cli


bbjubjub2494 (approved):
makes sense

PR #661, opened 2025-09-15
feat: removal of ackee-blockchain.solidity-tools and consensys.vscode-solidity-auditor

aldoborrero:


aldoborrero:
@bbjubjub2494 exactly, and less maintenance burden

bbjubjub2494 (dismissed):
Concept ACK. I would put a `throw` instead of deleting the entry so people know why it was removed.

selfuryon (dismissed):

selfuryon:
I just tried to add `throw` but `throw` logic breaks `platformApps` logic, need to exclude removed packages from the list of packages which passed to `platformApps` so it's just simpler to delete them :)

selfuryon (approved):

PR #700, opened 2025-10-17
chore(ci): try to migrate to auto-sign all commits by bot

selfuryon:
I want to have signed commits from the bot, but it's tricky. The core thing here is that we can archive that if we will use Github REST API to create commits instead of just `git commit` like it was done [here](https://github.com/peter-evans/create-pull-request/blob/46cdba753c74545733b821043d64bd6925fc4da9/src/create-pull-request.ts#L209)

So we have two ways to do that:
* adjust our `create-pr.sh` script to use `gh` instead of `git commit`
* switch to `create-pull-request` actions and just pass `sign-commits: true` there

I chose the 2nd one, I think it will be simpler. Just run an action manually and it works :)


aldoborrero:
If it works, I'm not against that ;)

So remove `update-pr.sh`.

PR #734, opened 2025-11-01
update flake-compat url

bbjubjub2494:
The nix-community one is no longer maintained https://github.com/nix-community/flake-compat


aldoborrero:
@bbjubjub2494 let's remove support for `flake-compat`. I added it a couple of years ago but I think these days it's not worthwhile (I know I'll piss some hardcore fans pure old nix).

PR #810, opened 2025-12-13
Extend start timeout so checkpoint sync can complete (nimbus)

ekimber:
The checkpoint sync in Nimbus runs as an ExecStartPre command. This is subject to the default unit timeout which is usually 90 seconds, although it may be set to more or less than that. I find it takes about 2 minutes to complete which means the checkpoint sync is aborted. Therefore I propose specifying a longer timeout in the unit.

Of course, YMMV depending on bandwidth etc but it seemed reasonable to go for 5 minutes.

### DIFF
--- a/modules/nimbus-beacon/default.nix
+++ b/modules/nimbus-beacon/default.nix
@@ -187,6 +187,7 @@ in {
                     then cfg.args.user
                     else user;
                   StateDirectory = user;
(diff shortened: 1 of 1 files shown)


bbjubjub2494:
> I find it takes about 2 minutes to complete

was this on Ethereum Mainnet? (I would expect that's the network that takes the most time)

PR #817, opened 2025-12-17
Add besu module

ekimber:
Added a besu module with basic configuration.


aldoborrero:
@ekimber Nice! Thanks!

PR #818, opened 2025-12-17
teku-beacon: init module

bbjubjub2494:
This is not complete, but has most common options. I've been able to run a Gnosis validator with it.


aldoborrero:
@bbjubjub2494 nice thanks! I see treefmt is complaining a little bit, can you run it?

PR #866, opened 2026-01-28
refactor(modules): migrate NixOS modules to RFC 42 settings pattern

selfuryon:
In favor to:
- [RFC 42: NixOS settings options](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md)
- [Discussion #325: Module System Consideration](https://github.com/nix-community/ethereum.nix/discussions/325)

https://github.com/nix-community/ethereum.nix/pull/847


aldoborrero:
@selfuryon I went ahead and created #870 and cherry-picked those two commits from this PR!

selfuryon:
@aldoborrero hey, thanks! Sorry, was busy with work but now returned to this PR :) 
I rebased it so it's ready I believe to take a look. I also wanted to add some additional modules but I will do that later to not delay it anymore

aldoborrero:
@bbjubjub2494 can you test this PR on your nimbus node if possible?

modules/nixos/nimbus-beacon/default.nix
```diff
@@ -180,26 +122,37 @@ in
...
+                lib.mkBefore [
+                  # Create keymanager token file if it doesn't exist
+                  ''
+                    ${pkgs.coreutils-full}/bin/cp --no-preserve=all --update=none \
+                    /proc/sys/kernel/random/uuid ${dataDir}/${keymanagerTokenFile}''
+                ]
+                ++ (optionals (trustedNodeUrl != null) [
+                  # Run checkpoint sync if trusted-node-url is configured
+                  "${cfg.package}/bin/${bin} trustedNodeSync ${checkpointSyncArgs}"
+                ]);
```
bbjubjub2494:
```suggestion
                lib.mkBefore ([
                  # Create keymanager token file if it doesn't exist
                  ''
                    ${pkgs.coreutils-full}/bin/cp --no-preserve=all --update=none \
                    /proc/sys/kernel/random/uuid ${dataDir}/${keymanagerTokenFile}''
                ]
                ++ (optionals (trustedNodeUrl != null) [
                  # Run checkpoint sync if trusted-node-url is configured
                  "${cfg.package}/bin/${bin} trustedNodeSync ${checkpointSyncArgs}"
                ]));
```

selfuryon:
Rebased onto current `main` and addressed the review feedback.

**@bbjubjub2494** — both your `nimbus-beacon` comments are fixed:

1. **`--network chiado` vs `--network=chiado`**: confirmed empirically against `nimbus_beacon_node` 26.5.0 — the space-separated form is rejected (`nimbus_beacon_node does not accept arguments`, because confutils treats the bare value as a positional), while `--network=chiado` parses fine. The module now generates args with `lib.cli.toCommandLine (name: { option = "--${name}"; sep = "="; explicitBool = false; })` instead of the (deprecated) `toGNUCommandLine`, so it emits `--key=value`. Verified the rendered `ExecStart`:
   `… nimbus_beacon_node_gnosis --data-dir=%S/… --network=chiado --rest --rest-port=5052`

2. **`ExecStartPre` / `mkBefore`**: adopted your suggestion — it was `lib.mkBefore [ … ] ++ (optionals …)`, i.e. `orderSet ++ list`, which fails module evaluation (`expected a list but found a set`). Wrapping the whole list in `mkBefore` fixes it; verified the service now evaluates and appends the `trustedNodeSync` step only when `trusted-node-url` is set.

Notes for reviewers:
- Other clients (geth/erigon/reth/besu/teku/lighthouse/prysm/mev-boost) still use `toGNUCommandLine` (space-separated), which they all accept — only Nimbus is strict. It's deprecated upstream though, so a follow-up could migrate them to `toCommandLine` too.
- `nimbus-validator` is not part of this refactor (still the old pattern) and is unaffected.

selfuryon:
Two follow-up cleanups pushed:

- **Dropped the deprecated `lib.cli.toGNUCommandLine`** everywhere — all modules now use `lib.cli.toCommandLine`. For the non-Nimbus clients this uses `sep = null` (space-separated), which produces byte-identical output to the old call (verified via eval for both the `--` and single-`-` bootnode variants), so no behavior change — just no more deprecation warnings.
- **Migrated `nimbus-validator` to the RFC 42 `settings` pattern** (it was still the old `network`/`extraArgs` module). Same approach as `nimbus-beacon`: `sep = "="`, gnosis binary for gnosis/chiado, metrics/keymanager firewall ports, and keymanager-token creation when the keymanager API is enabled.

Sanity-checked all 13 modules by evaluating each service's `ExecStart` under a NixOS config — all render correctly.

selfuryon:
Migrated the last remaining module, **helios**, to the RFC 42 `settings` pattern (it was still using a structured `args` submodule). The `network` key keeps selecting the subcommand (`ethereum` vs `opstack --network <chain>`), and socket activation / firewall read `rpc-bind-ip`/`rpc-port` from settings. Dropped `helios/args.nix`.

With this, the entire `modules/nixos/` tree is on the settings pattern — no `args.nix` files remain. Also added a note to AGENTS.md that Nimbus modules use `sep = "="` (confutils rejects space-separated values) while the other clients use `sep = null`.

Verified all module `ExecStart`s (and helios' subcommand/socket) render correctly via eval.

selfuryon:
@aldoborrero @bbjubjub2494 @brianmcgee can you review it pls?

bbjubjub2494 (approved):
very nice!

PR #944, opened 2026-02-20
ethrex: init at 21.0.0

aldoborrero:
## Summary
- Add ethrex (LambdaClass), a modular and ZK-native Ethereum execution client written in Rust
- Built with L1 and L2 rollup support (`l2` feature enabled)

## Notes
- Upstream `Cargo.lock` has crates (bls12_381, halo2curves-axiom, zkhash) from both crates.io and git forks with identical name-version; unified via `[patch.crates-io]`
- openvm crates appear from two git refs of the same repo; pinned to a single commit via `cargo update --precise` in the shipped Cargo.lock
- Vendor directory patched (`ln -sfT`) to handle remaining URL-only symlink collision between the two openvm source URLs
- `auditable = false` due to `cargo-auditable` incompatibility with Cargo edition 2024 feature syntax
- `VERGEN_IDEMPOTENT=1` skips git2 version detection

## Test plan
- [x] `nix build .#ethrex` succeeds
- [x] `ethrex --version` → `ethrex ethrex/v9.0.0-...`
- [x] `ethrex --help` shows `l2` subcommand

## Upstream
- Reported duplicate crate sources: https://github.com/lambdaclass/ethrex/issues/6234


ilitteri:
Hey! ethrex dev here, thank you for adding ethrex support for ethereum.nix. Status from our side, since this PR predates a lot of dependency churn:

- `9.0.0` is 12 major releases behind (latest is `v21.0.0`; we cut roughly every two weeks). A bump needs a regenerated `Cargo.lock` and new `outputHashes`: the `bls12_381` fork branch moved to `expose-affine-constructors`, `zisk` moved, and there are new git deps (`pil2-proofman`, a second stark-backend tag).
- The `cargo update --precise` unification of the two `openvm` refs no longer works: the guest pins `openvm` tag `v1.4.1` while the prover tracks a newer branch rev with different crate versions. The only remaining name+version collision across the two refs is `openvm-custom-insn` `0.1.0`, so the `ln -sfT` workaround shrinks to that one crate, or may be droppable.
- The February darwin buildbot log is gone (buildbot data reset in April), so only a rebase will show whether that failure persists.

selfuryon (approved):

PR #1235, opened 2026-07-21
chore: add versionCheckHook to core clients

selfuryon:
## Summary

Adds `versionCheckHook` install checks to the core execution and consensus clients so their built binaries are actually exercised (`--version`) during the build. This is cheap coverage that catches broken runtimes before a release ships.

Packages gaining the hook: **geth, reth, erigon, lighthouse, prysm, nimbus, teku**.

- `besu` already has an equivalent `testers.testVersion` check, so it was left as-is.
- `lodestar` is a `buildFHSEnv` wrapper which does not run the standard `installCheck` phase, so it needs a different approach and is out of scope here.

## Bug caught: teku was broken at runtime

Adding the hook to teku immediately surfaced that **teku 26.7.1 does not run at all** as currently packaged:

1. It is compiled for **Java 25**, but the package used the default `jre` (Java 21) → `UnsupportedClassVersionError` (`class file version 69.0` > `65.0`) on startup. Fixed by passing `jre = pkgs.jdk25` (matching `besu`).
2. The upstream gradle launcher script also needs `uname`/`xargs`/`sed` on `PATH`; these are now wrapped in via `coreutils`/`findutils`/`gnused`.

Without a run/version check, both defects were invisible.

## Testing

- `nix build .#geth` — passes with the version check.
- `nix build .#teku` — fails before the fix (both issues above), passes after.
- All 7 packages evaluate cleanly; `nix fmt` applied.
- Remaining clients (reth, erigon, lighthouse, prysm, nimbus) rely on self-contained `--version` output; left to CI to build-verify.

### DIFF
--- a/packages/erigon/package.nix
+++ b/packages/erigon/package.nix
@@ -3,6 +3,7 @@
   fetchFromGitHub,
   lib,
   nix-update-script,
+  versionCheckHook,
   subPackages ? [
     "cmd/erigon"
     "cmd/evm"
@@ -39,6 +40,9 @@ buildGoModule rec {
 
   ldflags = [ "-extldflags \"-Wl,--allow-multiple-definition\"" ];
   inherit subPackages;
+
+  doInstallCheck = true;
+  nativeInstallCheckInputs = [ versionCheckHook ];
 
   passthru = {
     category = "Execution Clients";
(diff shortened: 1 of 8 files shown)


bbjubjub2494 (approved):
nice catch!

PR #1239, opened 2026-07-23
modules: add user option and disable DynamicUser when set

selfuryon:
## What

Adds a `user` option to the NixOS modules that were missing it and makes setting a custom user actually run the service as a real, statically-managed user (not under systemd's `DynamicUser`).

- **Execution clients + others** (geth, erigon, besu, nethermind, reth, geth-bootnode, mev-boost, helios): new `user` option (nullOr str, default null). When unset, behaviour is unchanged — `DynamicUser = true` and `User = <service>-<name>`. When set, the service runs as that user.
- **All modules**: when `user` is set, `DynamicUser = mkForce false` is applied, since a real user is incompatible with `DynamicUser`. The consensus modules (prysm, lighthouse, teku, nimbus) already exposed `user` but kept `DynamicUser` on — they now disable it too.

## Why

This is the RFC 42 module equivalent of the request in #626, which targeted the old `modules/geth/` layout (removed in #866). That PR wanted to manage the geth service user directly instead of relying on `DynamicUser`. This applies the idea uniformly across all clients and supersedes #626.

## Testing

Evaluated the resulting systemd units for every touched module, with and without `user`:

- `user` set → `User = <user>`, `DynamicUser = false`
- `user` unset → `User = <service>-<name>`, `DynamicUser = true` (defaults preserved)

`nix fmt` clean.

### DIFF
--- a/modules/nixos/besu/default.nix
+++ b/modules/nixos/besu/default.nix
@@ -15,6 +15,7 @@ let
     mapAttrs
     mapAttrs'
     mapAttrsToList
+    mkForce
     mkIf
     mkMerge
     nameValuePair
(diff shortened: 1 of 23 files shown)


selfuryon:
@bbjubjub2494 @aldoborrero @brianmcgee pls review :)

bbjubjub2494 (approved):
I would personally favor a pattern where we `assert ((cfg.user != null) -> !(cfg.serviceConfig.DynamicUser or false))`; only users should be `mkForce`ing, and it's more pedagogical. I don't want to hold this up though so I'll approve it. Just ping me again if needed

thalin:
This would be really great to merge! I look forward to it.

PR #1252, opened 2026-07-26
lighthouse-validator: auto-set --unencrypted-http-transport when http enabled

### LINKED ISSUE #795: Lighthouse validator requires --unencrypted-http-transport to enable http
### Context

_No response_

### Description

The lighthouse validator client requires an `--unencrypted-http-transport` flag when enabling http access "to ensure the user is aware of the risks involved" so the `http.enable` option only works if you add that flag in extraArgs. It would be preferable if the extraArgs option wasn't explicitly required.

### Before submitting

- [x] I have verified that my idea is a change request and not a bug report

selfuryon:
## Description

Lighthouse refuses to serve the validator HTTP API over plain HTTP unless `--unencrypted-http-transport` is passed explicitly (to ensure the user is aware of the risks involved). Previously, users of the `lighthouse-validator` module had to add this flag manually via `extraArgs` whenever they set `settings.http = true`.

This change injects the flag automatically when `http` is enabled, unless the user already sets `unencrypted-http-transport` in `settings` (so no duplicate arg is produced).

## Testing

Evaluated the module for three instances and inspected the generated `ExecStart`:

| Case | `settings` | Result |
|------|-----------|--------|
| auto | `http = true` | `--http … --unencrypted-http-transport` (injected once) ✅ |
| explicit | `http = true; "unencrypted-http-transport" = true` | flag appears exactly once, no duplicate ✅ |
| nohttp | http unset | flag not present ✅ |

Ran `nix fmt`.

Closes #795

### DIFF
--- a/modules/nixos/lighthouse-validator/default.nix
+++ b/modules/nixos/lighthouse-validator/default.nix
@@ -94,6 +94,14 @@ in
           explicitBool = false;
         }) (processSettings normalSettings);
 
+        # Lighthouse refuses to serve the validator HTTP API over plain HTTP
+        # unless --unencrypted-http-transport is passed explicitly. Inject it
+        # automatically when http is enabled so users don't have to add it via
+        # extraArgs, unless they set the flag themselves in settings (#795).
+        httpTransportArgs = optionals ((s.http or false) && !(s ? "unencrypted-http-transport")) [
+          "--unencrypted-http-transport"
+        ];
+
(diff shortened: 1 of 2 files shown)


bbjubjub2494 (approved):
`extraArgs` can still introduce a duplicate argument right? Is it an issue to have duplicate args? If not, then this works, but it could be kept a bit more simple

PR #1253, opened 2026-07-26
vouch: add NixOS service module

### LINKED ISSUE #90: Add a service for Vouch

selfuryon:
## Description

Adds a NixOS service module for [Vouch](https://github.com/attestantio/vouch), Attestant's Ethereum multi-node validator client.

Unlike the flag-driven clients in this repo, Vouch is configured through a YAML file that it discovers via `--base-dir` (viper reads `<base-dir>/vouch.yaml`). The module therefore renders the freeform `settings` attrset to `vouch.yaml` using `pkgs.formats.yaml` and points Vouch at a directory containing it.

### Features

- Multiple named instances under `services.ethereum.vouch.<name>`.
- Freeform `settings` mirroring the [upstream configuration](https://github.com/attestantio/vouch/blob/master/docs/configuration.md) (nested YAML keys).
- Hardened systemd unit via the shared `baseServiceConfig` (DynamicUser, etc.).
- `openFirewall` opens the Prometheus metrics port, parsed from `settings.metrics.prometheus.listen-address`.
- `extraArgs` escape hatch.

Secrets should be provided as majordomo references (e.g. `file://...`) rather than inline, since the generated config lives in the world-readable Nix store — this is documented in the option description.

## Testing

Module is discovered by blueprint (`nix eval .#nixosModules` lists `vouch`). Evaluated a sample instance and verified:

- `ExecStart` = `…/bin/vouch --base-dir=/nix/store/…-vouch-main-base`
- Generated `vouch.yaml` matches the expected nested structure.
- `openFirewall` opens TCP `8081` (parsed from `127.0.0.1:8081`).
- `DynamicUser = true`, `StateDirectory = vouch-main`.

Ran `nix fmt`.

Closes #90

### DIFF
--- a/modules/nixos/vouch/options.nix
+++ b/modules/nixos/vouch/options.nix
@@ -1,0 +1,89 @@
+{
+  lib,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib)
+    mkEnableOption
+    mkOption
+    types
+    literalExpression
+    ;
+
+  vouchOpts = {
+    options = {
+      enable = mkEnableOption "Vouch, an Ethereum multi-node validator client from Attestant";
+
+      package = mkOption {
+        type = types.package;
+        default = pkgs.vouch;
+        defaultText = literalExpression "pkgs.vouch";
+        description = "Package to use for the Vouch binary.";
+      };
+
+      user = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = "User to run the systemd service. When null, a dynamic user is used.";
+      };
+
+      openFirewall = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Open the Prometheus metrics port in the firewall. The port is parsed
+          from `settings.metrics.prometheus.listen-address`.
+        '';
+      };
+
+      settings = mkOption {
+        type = types.submodule {
+          freeformType = (pkgs.formats.yaml { }).type;
+        };
+        default = { };
+        description = ''
+          Vouch configuration. Rendered verbatim to a `vouch.yaml` file that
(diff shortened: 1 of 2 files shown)


modules/nixos/vouch/options.nix
```diff
@@ -0,0 +1,89 @@
...
+          freeformType = (pkgs.formats.yaml { }).type;
+        };
+        default = { };
+        description = ''
+          Vouch configuration. Rendered verbatim to a `vouch.yaml` file that
+          Vouch reads through `--base-dir`. Use the nested keys documented at
+          <https://github.com/attestantio/vouch/blob/master/docs/configuration.md>.
+
+          Secrets (wallet passphrases, client certificates) should be provided
+          as majordomo references (e.g. `"file:///run/credentials/..."`) rather
```
bbjubjub2494, line 51: nitpick but what's the difference between a "majordomo reference" and a URL? What would a reader that's familiar with majordomo — which I am not, understand different than calling it a URL?

bbjubjub2494 (approved):
very nice that we can pass config files instead of messing with CLI arguments

modules/nixos/vouch/options.nix
thalin: Not really. A majordomo reference _is_ a URL, sometimes with a funky scheme. https://github.com/wealdtech/go-majordomo#usage

bbjubjub2494: Thanks! To answer my own question: the distinction is helpful because we know  for a majordomo reference we can use `direct` or `asm`, whereas a "URL" we don't know which schemes are expected

PR #1254, opened 2026-07-26
dirk: add NixOS service module

### LINKED ISSUE #89: Add a service for Dirk

selfuryon:
## Description

Adds a NixOS service module for [Dirk](https://github.com/attestantio/dirk), Attestant's distributed remote keymanager, focused on secure signing operations.

Like the companion Vouch module, Dirk is configured through a YAML file it discovers via `--base-dir` (viper reads `<base-dir>/dirk.yaml`). The module renders the freeform `settings` attrset to `dirk.yaml` using `pkgs.formats.yaml` and points Dirk at a directory containing it.

### Features

- Multiple named instances under `services.ethereum.dirk.<name>`.
- Freeform `settings` mirroring the [upstream configuration](https://github.com/attestantio/dirk/blob/master/docs/configuration.md) (nested YAML keys).
- Hardened systemd unit via the shared `baseServiceConfig` (DynamicUser, etc.).
- Persistent state + working directory at `/var/lib/dirk-<name>` for slashing protection and filesystem wallet stores.
- `openFirewall` opens the gRPC server port, parsed from `settings.server.listen-address`.
- `extraArgs` escape hatch.

Secrets (passphrases, keys) should be provided as majordomo references (e.g. `file://...`) rather than inline, since the generated config lives in the world-readable Nix store — documented in the option description.

## Testing

Module is discovered by blueprint (`nix eval .#nixosModules` lists `dirk`). Evaluated a sample instance and verified:

- `ExecStart` = `…/bin/dirk --base-dir=/nix/store/…-dirk-main-base`
- Generated `dirk.yaml` matches the expected nested structure (server, stores, peers, permissions).
- `openFirewall` opens TCP `13141` (parsed from `0.0.0.0:13141`).
- `DynamicUser = true`, `StateDirectory = dirk-main`, `WorkingDirectory = %S/dirk-main`.

Ran `nix fmt`.

Closes #89

### DIFF
--- a/modules/nixos/dirk/default.nix
+++ b/modules/nixos/dirk/default.nix
@@ -1,0 +1,100 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  modulesLib = import ../../../lib/modules.nix lib;
+
+  inherit (lib)
+    concatStringsSep
+    filterAttrs
+    flatten
+    last
+    mapAttrs'
+    mapAttrsToList
+    mkIf
+    mkMerge
+    nameValuePair
+    optionals
+    splitString
+    toInt
+    ;
+  inherit (lib.attrsets) zipAttrsWith;
+  inherit (modulesLib) baseServiceConfig;
+
+  eachDirk = config.services.ethereum.dirk;
+
+  format = pkgs.formats.yaml { };
+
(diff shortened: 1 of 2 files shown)


bbjubjub2494 (approved):
same comments as with #1253

